PR #88, opened 2014-11-21
Add QtWebEngine ebuild

yuri-sevatz:
My attempt to get QtWebEngine working on Gentoo.

This *DEFINITELY* doesn't work in all cases now... but I'm pretty sure this is supposed to be in the tree.

My latest build failure seems to revolve around PYTHON 2 print statements not working when gentoo is configured to use python 3 as the default interpreter.  Not sure how to switch the package's build to use 2_x.

Would appreciate any help on this, I really want to start writing code for this =)


chatellier:
Plz merge it. I also need this ebuild. Thx

Pesa:
I don't think this will go in any time soon. QtWebEngine is a mess of bundled libraries and a maintenance nightmare. Debian and Fedora maintainers are refusing to add this package to their repos as well, see e.g. https://marc.info/?l=kde-core-devel&m=142954900813235&w=2

yuri-sevatz:
I agree with the criticism,.  Dynamic linking would be desirable.  It's a pretty big build.  Lowest I tried it on was with 8GB, though I didn't notice anything as off-putting like the other distros are complaining about with just a standard gold build.

For early adopters, I'll try and maintain it in my fork until it gets a home, or until there's more control over static/dynamic linking.  It's pretty decoupled from everything.  And it's very lightweight at runtime from what I've seen.

Also -- w.r.t  WebChannel: We can probably talk about a separate merge for that, since that uses mostly just QObject introspection.  There is a dependency in WebEngine on WebChannel in 5.5, but otherwise it's 100% decoupled from WebEngine.

yuri-sevatz:
@chatellier - Copy the ebuild from my repository to a local overlay in the meantime:

https://wiki.gentoo.org/wiki/Overlay/Local_overlay

Hopefully that should unblock anything you're trying locally.

ff2000:
Qt 5.6 Alpha seems to make it easy to use system libs instead of the bundled ones for webengine. At least that's in the ChangeLog.

Pesa:
I know, there are more details here: http://lists.qt-project.org/pipermail/development/2015-August/022869.html

kensington:
Thanks, an initial version of dev-qt/qtwebengine is in the overlay now.

PR #89, opened 2015-01-19
[lxqt-base] LXQt SRC_URI change

jubalh:
On 18.01.2015 J. Leclanche announced on the LXDE mailing list that the
package download path will change in a few weeks (with the release of
0.9.0). The change of download paths affacts all version.

The new locations are already working, so here is the change that needs
to be done.


jubalh:
for cross reference: https://bugs.gentoo.org/show_bug.cgi?id=537032

kensington:
Thanks, merged. Sorry about the delay.

jauhien:
Sorry about the delay, I had no access to computer on which I could test, merge to CVS.

@kensington: thank you.

PR #90, opened 2015-02-21
[RFC] Add multilib sni-qt

karolherbst:
This pull request adds multilib support for sni-qt fixing these gentoo bugs:
https://bugs.gentoo.org/show_bug.cgi?id=516488
https://bugs.gentoo.org/show_bug.cgi?id=530154

The commits basically copy ebuilds from upstream gentoo and then add multilib support in needed packages:
* qjson
* libdbusmenu-qt (here I splitted the ebuild into slots, because only the qt4 part needs to be multilib and it would messy to exclude it for the qt5 multibuild variant if it would be in one ebuild, no conflicts detected)
* sni-qt

I've tested it with skype inside plasma5 and the systray icons shows up.


karolherbst:
I think I should make this request in the kde overlay, because it belongs there more

karolherbst:
for reference: https://github.com/gentoo/kde/pull/161

PR #91, opened 2015-03-07
[dev-qt/qtwebkit23] add support for building with libxml2[icu]

enricotagliavini:
Add a patch to let it build when libxml2 is compiled with icu support
enabled. This solves conflicts with, e.g., chromium depending on
libxml2[icu]

I tested this ebuild to see if it is solving bug #522880 (when phonon-gstreamer-4.8.x is used there is a gstreamer 0.10, loaded by qtwebkit, vs 1.x, loaded by phonon, symbol collision in applications needing both gstreamer and qtwebkit, like amarok). The issue is solved as expected and phonon-gstreamer 4.8 works as expected. See also comment #35 in bug #522880

Note this ebuild currently generates a file collision with dev-qt/qtdeclarative-4.8.x
```
 * Detected file collision(s):
 *
 *      /usr/lib64/qt4/imports/QtWebKit/libqmlwebkitplugin.so
 *      /usr/lib64/qt4/imports/QtWebKit/qmldir
 *
 * Searching all installed packages for file collisions...
 *
 * Press Ctrl-C to Stop
 *
 * dev-qt/qtdeclarative-4.8.5:4::gentoo
 *      /usr/lib64/qt4/imports/QtWebKit/libqmlwebkitplugin.so
 *      /usr/lib64/qt4/imports/QtWebKit/qmldir
```
So i forcefully disabled webkit support in qtdeclarative by passing -no-webkit and dropped the DEP. Packages depending on qtdeclarative[webkit], like kdevelop, will need a little adjustment. With this change qtwebkit 2.3.4 can be installed, I fired up rekonq briefly and it worked as expected (tried to load youtube in html5).

### DIFF
--- a/dev-qt/qtwebkit23/files/qtwebkit23-2.3.4-use-correct-typedef.patch
+++ b/dev-qt/qtwebkit23/files/qtwebkit23-2.3.4-use-correct-typedef.patch
@@ -1,0 +1,40 @@
+From 916f00008b602ae1b260106e7fb1274d2282f61f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Jan=20Kundr=C3=A1t?= <jkt@flaska.net>
+Date: Tue, 3 Sep 2013 16:59:35 +0200
+Subject: [PATCH] ICU has defined UChar32 to be an int32_t since 2002
+
+This fixes the build failure of qtwebkit23 on my Gentoo machine.
+---
+ Source/WTF/wtf/unicode/qt4/UnicodeQt4.h     |    2 +-
+ Source/WTF/wtf/unicode/wchar/UnicodeWchar.h |    2 +-
+ 2 files changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/Source/WTF/wtf/unicode/qt4/UnicodeQt4.h b/Source/WTF/wtf/unicode/qt4/UnicodeQt4.h
+index a2d1ad4..392d2db 100644
+--- a/Source/WTF/wtf/unicode/qt4/UnicodeQt4.h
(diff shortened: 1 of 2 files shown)


Pesa:
I submitted a similar patch against the 4.8 branch of qt some time ago, but upstream rejected it.

enricotagliavini:
That could even be your patch, I took it from the gentoo bugzilla, Kensington pointed me to that. That said I looked a bit into libxml2 code, to me that piece of code (the original) doesn't look right and likely is a leftover / forgotten piece. Can be wrong, I'm not into that stuff, I started looking at it just because the problem we are having in gentoo with phonon-gstreamer is really annoying me. I don't like non upstreamed patches, but the qtwebkit situation is a bit messy and will sealf heal with qt5 anyway since icu support seems to be fixed there (and well there is no more webkit anyway in qt, it will be qtwebengine). So at least we can remove some annoyance to the gentoo users while waiting qt4 to be obsolete.

An alternative, but equally not very nice solution, is to disable icu support in libxml2 and have all packages needing libxml2 with icu support enabled to bundle it (like chromium), but between the two I prefer the not upstream patch by a far margin.

Pesa:
No, my patch was for the webkit copy in the 4.8 branch of qt. This is for the 2.3 branch of the "standalone" qtwebkit, aka qtwebkit-23 (yeah it's very confusing, blame upstream).

My patch, FTR, https://codereview.qt-project.org/28797

kensington:
I don't really understand why it was rejected upstream.

kensington:
Thanks, rebased and merged. I don't see any reason right now to not use the patch.

enricotagliavini:
Thank you very much

PR #92, opened 2015-03-18
Make it possible to configure debug builds on a per-package basis

jktjkt:
Previous version of this code only did something useful for qtbase, and
relied on qmake to propagate this change to other Qt modules. Because Qt
modules can be easily built with different debug settings, let's make it
work.

Tested with USE=-debug on everything and USE=debug on
qtdeclarative-5.5.9999.

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -658,6 +658,7 @@ qt5_qmake() {
 		QMAKE_LFLAGS="${LDFLAGS}" \
 		QMAKE_LFLAGS_RELEASE= \
 		QMAKE_LFLAGS_DEBUG= \
+		$(use debug && echo "CONFIG+=debug" || echo "CONFIG-=debug") \
 		"${projectdir}" \
 		"$@" \
 		|| die "qmake failed (${projectdir})"


Pesa:
While this *might* work, I don't want to support it officially because upstream has been very clear about the fact that toggling release/debug in qt modules can break the ABI.

If you want debug symbols you can do that quite easily in gentoo by configuring per-package `CXXFLAGS` in `/etc/portage`.

jktjkt:
Where did they say that?

Pesa:
Quite some time ago on IRC, I specifically asked them.

wrt the debug symbols, ok, I was just saying in case they were sufficient for your needs.

jktjkt:
As far as I know, the ABI is actually stable, at least on Linux. Who provided you with that information, and what exactly did they say? Please note that on e.g. Windows, the ABIs are indeed different, so your question might have been interpreted inaccurately ("can I toggle between debug and release? - nope, it won't work on Windows").

Pesa:
I'm pretty sure I introduced myself as a Linux distro packager... anyway, you're encouraged to ask again and report back ;) As I said, this happened a long time ago (~2 years I think) and I can't remember all the details. It's also possible that things have changed in the meantime and there are more guarantees now.

jktjkt:
This is fully expected to work well. This response is from a QtWebkit developer, IIRC: http://lists.qt-project.org/pipermail/interest/2015-March/015993.html

Pesa:
That answer focuses on debug symbols, which IIUC are not what you're interested in.

But maybe the ABI problem was fixed some time ago... http://code.qt.io/cgit/qt/qtbase.git/commit/?id=9b378143a8e8e309621d79ebed421f15dee0039e

jktjkt:
Right, and thanks for the link -- that looks like accidental breakage which is now fixed.

Fortunately, Thiago's response at http://lists.qt-project.org/pipermail/interest/2015-March/015994.html cleared any doubt here.

This commit to the eclass is not enough of course due to the [debug=] USE deps -- my bad.

Pesa:
There's also the question of what we pass to `./configure`... always using `-release` there seems the safest option, and then we override via `CONFIG`. Does it make sense?

jktjkt:
I haven't checked whether `qtbase` respects `CONFIG+=foo` syntax. My guess is that it won't because you have to bootstrap `qmake` as a part of the initial `qtbase` build, so I don't see what would take a look at the `CONFIG` values.

The safest option, IMHO, is to always pass either `CONFIG+=debug` or `CONFIG+=release` in all packages but `qtbase`, and rely on `configure`'s options for `qtbase`. The downside of this is that whatever you use for `qtbase` becomes a default, but I don't think there's any sane way to fix this.

Pesa:
Do you really mean `qtbase` (i.e. all modules in it), or just the `qtcore` package?

jktjkt:
I meant "whatever gets build as a very first part of Qt5", which is the entire `qtbase` in upstream's default, and `qtcore` in Gentoo. The point is that I suspect that the release vs. debug settings of this part affect the defaults for the rest of the stack.

Pesa:
> I meant "whatever gets build as a very first part of Qt5", which is the entire qtbase in upstream's default, and qtcore in Gentoo.

I haven't actually tried yet so I might be wrong, but I think only the initial qmake bootstrap (part of `qtcore`) cannot be handled via `CONFIG`, since qmake is not used at all in that case. Everything else does use qmake, so passing `CONFIG` should be enough. Or am I missing something?

> The point is that I suspect that the release vs. debug settings of this part affect the defaults for the rest of the stack.

Correct. That's why I suggested to default to a release build, in case we miss overriding something...

PR #93, opened 2015-03-22
[dev-qt/qtgui] enable threaded mesa in xcb-plugin

ff2000:
http://cgit.freedesktop.org/xcb/libxcb/commit/?id=be0fe56c3bcad5124dcc6c47a2fad01acd16f71a
was commited 2014-03-30 and released in August 2014 with libxcb-1.11,
which is already stable in Gentoo.
Enabling threaded Mesa should make gui applications more responsive in
general.


ff2000:
Some notes:
I did NOT try to emerge the live ebuilds. I changed the patch so that it fits the current state of the branches. So have a good look before merging.
I am running kde with this patch enabled since about a week. No new issues came up. But it fixes some issues: Occasionally a drag was started though I never clicked. This happend regularly on the lock screen, kickoff and plasma-volume-control. This completely disappeared.
Also plasma flickered quite often for me, e.g. when closing a systray popup or opening a new window. I did not see that since using this patch.

Users of proprietary drivers (AMD/NVIDIA) are not affected, but nouveau and intel should benefit quite alot.

Pesa:
Shouldn't this go upstream?

ff2000:
I just opened upstream issue:
https://bugreports.qt.io/browse/QTBUG-45170
I just don't know if it is possible. Not every downstream (-> Distros) already ships libxcb-1.11 or will ship in the near future. And I think Qt devs know about the workarounds they introduced, so they will fix this issue when they think it's time.

(BTW.: the qt issue tracker sucks :( Is it possible to just get a plain old-fassioned non-JS form to add new issues? I use vimperator and I need to escape from INSERT in order to navigate buffers, which is impossible with that tracker - moving focus out of the input fields will empty the form...)

Pesa:
Thanks a lot. It's definitely doable, the configure script can check for the libxcb version and blacklist only if it's older than 1.11

ff2000:
Three weeks passed and no comment from upstream...
If you think it is good to switch to threaded mesa in Gentoo I will rebase.

Pesa:
I cannot make an informed decision about this, as I don't know all the implications of this change. But we can definitely try it for 5.5, and revert if it causes problems. So please rebase the commit and amend it so that it touches only `5.5.9999` and `5.9999`. Thanks.

Pesa:
ping?

ff2000:
Sorry for the late response.
I will close this for now. I actually can't test if the live ebuilds work/patches apply. I also don't use any Qt app since quite a while and I don't have that much time left.
Probably someone else wants to jump in, or I have use (and time) for it in the future.

PR #95, opened 2015-04-24
Add QtWebChannel ebuild

yuri-sevatz:
Isolated pull request with just WebChannel.

This is basically just for serializing access to QObjects over streams.


Pesa:
Do not include a non-live 5.4.x ebuild. I will add 5.4.1 directly to the main tree once the live ebuilds are merged here.

Pesa:
I went ahead, did the necessary changes, and pushed the live ebuilds to the overlay. Thanks!

PR #96, opened 2015-05-21
Minitube

yuri-sevatz:
Fix SRC_URI, add MINITUBE_GOOGLE_API_KEY from portage downstream.

Also, if you're testing these compile-time keys, pay special attention to this... http://flavio.tordini.org/forums/topic/error-downloading-server-reply-forbidden#post-93714

Google is doing a real number on their APIs right now, but this seems to work.


kensington:
Thanks!

PR #97, opened 2015-05-22
[app-admin/keepassx] Remove QtTest if test not set

anders-larsson:
Building app-admin/keepassx-9999 currently fails bcause dev-qt/qttest:4 isn't pulled in as a dependency.

Error message (versions mentioned in error message seems to refer to some other qt package. Perhaps dev-qt/qtcore).
``````
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:138 (message):
  Could NOT find Qt4 (missing: QT_QTTEST_INCLUDE_DIR QT_QTTEST_LIBRARY)
  (found suitable version "4.8.6", minimum required is "4.6.0")
``````

### DIFF
--- a/app-admin/keepassx/keepassx-9999.ebuild
+++ b/app-admin/keepassx/keepassx-9999.ebuild
@@ -27,6 +27,11 @@ DEPEND="${RDEPEND}
 	test? ( dev-qt/qttest:4 )
 "
 
+src_prepare() {
+	 use test || \
+		 sed -e "/^set(QT_REQUIRED_MODULES/s/QtTest//" -i CMakeLists.txt || die
+}
+
 src_configure() {
 	local mycmakeargs=(
 		$(cmake-utils_use_with test TESTS)


kensington:
It's already pulled in DEPEND behind the test USE flag, but CMakeLists.txt tries to find it unconditionally.

A solution might look something like:
    src_prepare() {
        use test || sed -e "/^set(QT_REQUIRED_MODULES/s/QtTest//" -i CMakeLists.txt || die
    }

anders-larsson:
dev-qttest:4 did not get pulled in as a dependency on my system. I did have dev-qttest:5 installed though but that should not satisfy this dependency. It was only if I merged dev-qt/qttest:4 manually (or after modifying the ebuild) that app-admin/keepassx-9999 could be merged successfully.

Perhaps I'm missing something but modifying CMakeLists wouldn't help in this case?

kensington:
The problem is the package looks for QtTest unconditionally - but it's really only required for tests. Thus, we should patch the package appropriately - otherwise we force everyone to install QtTest whether they run tests or not.

anders-larsson:
Ah. Now I understand. Of course it's preferable to not have dev-qt/qttest:4 as a hard dependency. I've updated the commit according to your instructions.

anders-larsson:
This pull request is not relevant any more since keepassx-9999 requires qt5 now. Will send a new pull request.

PR #102, opened 2015-06-11
fix silly missing directory bug (bug #551676 in Gentoo bugzilla)

sarnold:
I guess I'm too far out of the loop, since some of this stuff is no longer in my CVS(?) so here's a fix for silly bug #551676.


Pesa:
Thanks. For now I simply removed the `|| die`... but I will refactor this code soon because I don't like it in its current form.

PR #104, opened 2015-07-10
Update from repo

alexws54tk:
Gentoo does ssl by default now


Pesa:
Changed the commit message and applied. Thanks.

PR #106, opened 2015-09-17
[app-admin/keepassx] Update ebuild for qt5

anders-larsson:
keepassx can only be built with qt5 now.


anders-larsson:
Thanks for your feedback. I've updated the commit. Dependencies has been split between DEPEND and RDEPEND.

kensington:
Thanks, merged.

PR #107, opened 2015-09-23
x11-misc/sddm: sync live ebuild with tree

jfmcarreira:
Update live ebuild to cope with changes in the version 0.12.0


kensington:
Thanks, rebased as 9f3b848a1fc10768f6f55f0c523c8e039acd8723 with some minor fixes.

PR #108, opened 2015-09-26
QtBluetooth requires QtNetwork

dflogeras:


Pesa:
Slightly amended and merged. Thanks.

PR #109, opened 2015-09-27
dev-qt/qtmultimedia-5.5.0 Enabling gstreamer1.0 and let gstreamer0.10…

waltercool:
I tried compiling under gstreamer 1.0 and gstreamer 0.10 and works good. Gstreamer1.0 under QtMultimedia and QtWebkit using USE multimedia use flag works good (tested with QupZilla with videos).

Kwin from gentoo upstream seems to not like really this change (expect gstreamer USE flag and not gstreamer010), but I can confirm using gstreamer1.0 works very good (USE flag  gstreamer as supported by Kwin)

### DIFF
--- a/dev-qt/qtmultimedia/metadata.xml
+++ b/dev-qt/qtmultimedia/metadata.xml
@@ -7,6 +7,8 @@
 			inside the event loop (recommended by upstream)</flag>
 		<flag name="qml">Build QML/QtQuick bindings and imports</flag>
 		<flag name="widgets">Build the QtMultimediaWidgets module</flag>
+		<flag name="gstreamer">Enable HTML5 audio/video support via <pkg>media-libs/gstreamer:1.0</pkg></flag>
+		<flag name="gstreamer010">Enable HTML5 audio/video support via <pkg>media-libs/gstreamer:0.10</pkg></flag>
 	</use>
 	<upstream>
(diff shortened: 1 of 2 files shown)


waltercool:
I have ammended the last commit with the metadata modification.

waltercool:
Sorry to mess too much with travis, I had a git issue.

Pesa:
Do we really need to keep 0.10 support?

waltercool:
I don't really think so, we can move straight to gstreamer 1.0, but QtMultimedia supports both. This decision will depend of Gentoo politics about supporting legacy libraries. Because QtMultimedia acts as a framework for gstreamer and other libraries, we can easily move straight to 1.0 and drop 0.10 support.

If you think we should also drop gstreamer 0.10 support to gain gstreamer 1.0, notify me and I will remove the support from there.

kensington:
I'd prefer to keep 0.10 at least for now (last time I tested, there was still some potential symbol collision issues with 1.0 forced).

Pesa:
Fair enough.

kensington:
Thanks, applied in dd4deb2c3974d6b388040c78e78bb0213e5fbecf.

Vindex17:
Good, I wanted to get rid of gstreamer 0.10 too

PR #110, opened 2015-09-28
lxqt-base/lxqt-panel: fix dependencies

rysiowski:


Pesa:
Thanks.

Delete the first line of the commit message and leave only "lxqt-base/lxqt-panel: fix dependencies"

Pesa:
Amended and merged, thanks!

PR #112, opened 2015-09-29
Fix for building on non-glibc (musl/uclibc).

dflogeras:
This patch is being accepted by upstream for the 5.6 branch, but will
not be backported.  Therefore I have created these patches for the 5.5 branch.


Pesa:
Prepend `dev-qt/qtnetwork: ` to the commit message please, and mention the upstream commit SHA somewhere.

dflogeras:
Patch is in their staging hopper, once it is merged I will rebase with requested comments.

Pesa:
Amended and merged. Thanks!

PR #113, opened 2015-10-16
Eclass: fix LD variable

Vindex17:
Why is the linker set as the c++ compiler?


Pesa:
NAK

This patch is clearly wrong. qmake uses `LD` directly in the linking step of the generated Makefiles. But the linker is rarely invoked directly, the compiler invokes it after parsing the arguments.

PR #114, opened 2015-11-01
eclass: switch the default toolkit to qt5

Vindex17:


Pesa:
Thanks, but it seems a bit too early for this. I don't think every package in the tree is ready, and as the TODO in the code says, we need an eselect module. Are there any distros that have qt5 as the default version?

Pesa:
You should open a bug in bugzilla to track the progress towards this goal, apparently we don't have one yet.

Vindex17:
I'm aware of Arch Linux, not sure about openSUSE, Fedora and (K)Ubuntu

PR #118, opened 2015-12-22
Pcmanfm qt

hashashini:
pcmanfm-qt-9999.ebuild won't build any longer as libfm-qt was split from pcmanfm-qt upstream. This PR adds new ebuild libfm-qt-9999 and adjusts pcmanfm-qt-9999.ebuild so that it works again.
As I couldn't find the string "LIBRARY DESTINATION" anywhere in libfm-qt's CMakeLists.txt I removed the src_prepare part from pcmanfm-qt-9999.ebuild entirely.


Pesa:
The history you pushed is quite messy... there are several pairs of commit+revert... can you clean it up and push a single commit? Thanks.

PR #120, opened 2016-02-19
Remove openvg from default qt4 set

gh2k:
This isn't really useful on desktop, and can cause issues with mesa as it pulls in egl support

Given that these features are only really useful on embedded, it's best to leave this out of the default set and have the user pull in manually if desired


Pesa:
makes sense

Pesa:
I've reformatted the commit message and merged. Thanks for your contribution!

PR #122, opened 2016-04-18
dev-qt/qtwebengine: Fix 5.6.0 build with system-nss

a17r:
Patches taken from upstream.


Pesa:
I assume this fixes bug 577676..?

a17r:
Yes, I just found that as well

Pesa:
Then add it to the commit message please. And also mention that the patches come from upstream.

a17r:
done!

Pesa:
LGTM

Pesa:
Thanks!

PR #123, opened 2016-04-28
www-client/qupzilla: Bump EAPI, description, and more.

Chiitoo:
This set of changes includes the following:

- Update PLOCALES.
- Remove unneeded patch.
- Bump to EAPI 6.
- Update description.

Each change comes with its own commit.  I was not sure if it would be preferred to have them *all for one*, but in the end, I went for *one for all*.


Pesa:
Thanks!

PR #124, opened 2016-05-20
www-client/qupzilla: Fix build, bump plugins, and add PLOCALES.

Chiitoo:
These changes do the following:

- Fix install phase failure due to a renamed file upstream.
- Update plugins and PLOCALES.
- And finally, remove a whitespace that I don't see a reason for in 'SRC_URI+'.


www-client/qupzilla/qupzilla-9999.ebuild
```diff
@@ -27,7 +27,7 @@ fi
 
 DESCRIPTION="A cross-platform web browser using QtWebEngine"
 HOMEPAGE="http://www.qupzilla.com/"
-SRC_URI+=" https://github.com/QupZilla/${PN}-plugins/archive/${PLUGINS_HASH}.tar.gz -> ${PN}-plugins-${PLUGINS_VERSION}.tar.gz"
+SRC_URI+="https://github.com/QupZilla/${PN}-plugins/archive/${PLUGINS_HASH}.tar.gz -> ${PN}-plugins-${PLUGINS_VERSION}.tar.gz"
```
Pesa: don't we need to separate this URI from the previous one with a space?

Chiitoo: Ahhh, crud, I forgot to test this with the 2.0.0 ebuild!

Yeah, without the space, 2.0.0 will trip and fall over without the space.  It looked very weird to me, as I don't recall seeing it before.  I did imagine it had a reason!  :]

Unfortunately, no one was around IRC to reply when I asked about this, and I apparently sort of partially forgot to do more testing... me bad.

So, going by what I've seen before, would it make sense changing the 'SRC_URI+' part into something like this:
```
SRC_URI="${SRC_URI}
        https://github.com/QupZilla/${PN}-plugins/archive/${PLUGINS_HASH}.tar.gz -> ${PN}-plugins-${PLUGINS_VERSION}.tar.gz"
```

I'm fine with leaving it with the space, too, though.  Just curious more than anything.  ^^

Pesa: Absolutely. `foo+=" bar"` and `foo="${foo} bar"` are completely equivalent, except that the latter is more portable.

Chiitoo: As I thought, thanks.

Should I drop the change from this pull request though, since it's not exactly fixing anything for the time being?

Pesa:
Dropped patch 4/4, rebased, merged, pushed. Thanks ;-)

Chiitoo:
Cool!

Thanks.  :]

PR #125, opened 2016-05-30
media-video/obs-studio: New package.

Chiitoo:
Base for the ebuild found from https://github.com/saintdev/obs-studio-overlay.

Changes I made:

- Added support for ALSA.
- Removed qt5 USE-flag and related conditionals.  While the graphical user interface can be disabled, I have not been able to figure out a use for the package without it (all it seems to install then are header files, translations, and libraries, but no executable binaries).  By default, the UI is not disabled, so I took the cmake arg out as well.
- Added 'EGIT_SUBMODULES=()' to skip git cloning repositories seemingly specific to OS X and Windows.

I also looked into removing translations not defined by the user, but I'm not sure of the best way to handle it (I experimented with using 'find' and the 'l10n' eclass).

I'm also wondering about a way (aside from grepping glxinfo) to make sure the system does at least OpenGL 3.2, which obs-studio requires.

Any comments appreciated!


media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+
+EAPI="6"
+
+inherit cmake-utils eutils
+
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3
+	EGIT_SUBMODULES=()
+	EGIT_REPO_URI="https://github.com/jp9000/obs-studio.git
+		git://github.com/jp9000/obs-studio.git"
```
Pesa: the `https:` URI is enough

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+else
+	SRC_URI="https://github.com/jp9000/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~x86 ~amd64"
+	RESTRICT="primaryuri"
```
Pesa: why?

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+inherit cmake-utils eutils
+
+if [[ ${PV} == *9999 ]]; then
+	inherit git-r3
+	EGIT_SUBMODULES=()
+	EGIT_REPO_URI="https://github.com/jp9000/obs-studio.git
+		git://github.com/jp9000/obs-studio.git"
+else
+	SRC_URI="https://github.com/jp9000/${PN}/archive/${PV}.tar.gz -> ${P}.tar.gz"
+	KEYWORDS="~x86 ~amd64"
```
Pesa: sort alphabetically

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	dev-qt/qtdeclarative:5
+	dev-qt/qtgui:5
+	dev-qt/qtmultimedia:5
+	dev-qt/qtnetwork:5
+	dev-qt/qtquickcontrols:5
+	dev-qt/qtsql:5
+	dev-qt/qttest:5
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5
+	media-libs/x264
```
Pesa: A `:=` slot operator would probably make sense

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	media-video/ffmpeg
```
Pesa: ditto

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	x11-libs/libXinerama
+	x11-libs/libXcomposite
+	x11-libs/libXrandr
+	alsa? ( media-libs/alsa-lib )
+	fdk? ( media-libs/fdk-aac )
```
Pesa: ditto

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,86 @@
...
+	imagemagick? ( media-gfx/imagemagick )
+	jack? ( virtual/jack )
+	pulseaudio? ( media-sound/pulseaudio )
+	truetype? (
+		media-libs/fontconfig
+		media-libs/freetype
+	)
+	v4l? ( media-libs/libv4l )
+"
+
```
Pesa: The *de facto* standard is to leave no space between `DEPEND` and `RDEPEND`

Pesa:
Looks pretty good overall. I assume you're going to maintain it through a proxy maintainer? Any reasons you didn't send the PR to the main gentoo repo?

Chiitoo:
Thought so about the https URI, but I have no idea about the 'primaryuri' restriction.  Perhaps the original creator(s) wanted to disable needlessly checking the GENTOO_MIRRORS first?

Everything else noted too.

As for why not the main repo, I'm not entirely sure, but it might have made more sense!  I did ask kensington once upon a time if the qt-overlay would make sense for this, and there were no objections at least.

I believe I should be able to proxy maintain this for the time being.

Chiitoo:
So having made the changes already when we last conversed here, is there something else that should be done still, and/or should this really go in the main repo instead?

Thanks!

media-video/obs-studio/metadata.xml
```diff
@@ -0,0 +1,26 @@
...
+    internally.
+  </longdescription>
+  <use>
+    <flag name="alsa">Enable the use of ALSA for audio recording.</flag>
+    <flag name="fdk">Enable libfdk support for AAC encoding.</flag>
+    <flag name="imagemagick">Use ImageMagick for image loading instead of FFmpeg.</flag>
+    <flag name="jack">Add support for the JACK Audio Connection Kit.</flag>
+    <flag name="pulseaudio">Enable the use of PulseAudio for audio recording.</flag>
+    <flag name="truetype">Enable the Freetype text plugin.</flag>
+    <flag name="v4l">Enable the v4l2 plugin.</flag>
```
Pesa: I'd remove all local USE descriptions except for `fdk` and `imagemagick`, since they're all very similar to the global descriptions.

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,83 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI="6"
+
+inherit cmake-utils eutils
```
Pesa: where is `eutils` used?

media-video/obs-studio/obs-studio-0.14.2.ebuild
```diff
@@ -0,0 +1,83 @@
...
+	dev-qt/qtwidgets:5
+	dev-qt/qtx11extras:5
+	media-libs/x264:=
+	media-video/ffmpeg:=
+	x11-libs/libXinerama
+	x11-libs/libXcomposite
+	x11-libs/libXrandr
+	alsa? ( media-libs/alsa-lib )
+	fdk? ( media-libs/fdk-aac:= )
+	imagemagick? ( media-gfx/imagemagick )
```
Pesa: `media-gfx/imagemagick` also has a subslot, but I dunno if this package links to it or just uses the command-line binaries. Please check whether a `:=` operator should be added to this dependency.

Pesa:
> should this really go in the main repo instead?

The software doesn't seem too qt-centric, it just happens to use qt... so I think you should contact proxy-maint and have it merged into gentoo.git. However I don't have a particularly strong opinion about this, and I'm not opposed to merging it into the qt overlay. @kensington?

Chiitoo:
USE-descriptions changed.  Was indeed wondering if I should simply leave out the ones that are globally accurate enough...

Not sure about 'eutils'.  Possibly a remnant of old that I didn't really think of thinking of.  Removed as well.

I don't know if this is a definite way of checking, but:
```
$ lddtree /usr/bin/obs
obs => /usr/bin/obs (interpreter => /lib64/ld-linux-x86-64.so.2)
    libobs.so.0 => /usr/lib64/libobs.so.0
        libdbus-1.so.3 => /usr/lib64/libdbus-1.so.3
        libMagickCore-6.Q16.so.2 => /usr/lib64/libMagickCore-6.Q16.so.2
[...]
```
I've added the sub-slot for imagemagick as well.

Thanks again!

kensington:
> The software doesn't seem too qt-centric, it just happens to use qt... so I think you should contact proxy-maint and have it merged into gentoo.git. However I don't have a particularly strong opinion about this, and I'm not opposed to merging it into the qt overlay. @kensington?

I agree. If the package is of a decent quality gentoo.git probably is a better home, but as the person who would be taking care of it that'd be ultimately up to you, @Chiitoo.

Pesa:
> That, and the main repo seeming a bit more scary than this here overlay.

Nah, don't worry, this ebuild looks of much better quality than several others I've seen in gentoo.git ;)

Pesa:
> I don't know if this is a definite way of checking, but:
> [...]

yeah it definitely links to the imagemagick library.

Chiitoo:
Done did it, and opened up a request over at gentoo/gentoo#1690.

Thanks a bunch!  :]

PR #128, opened 2016-07-07
[dev-qt/qtdatavis3d] New package.

tbrethom:
Please feel free to comment.

I'm not 100% sure about the gles2 part.


dev-qt/qtdatavis3d/qtdatavis3d-5.7.0.ebuild
```diff
@@ -0,0 +1,29 @@
...
+DEPEND="
+	~dev-qt/qtcore-${PV}
+	~dev-qt/qtdeclarative-${PV}[gles2=]
+	~dev-qt/qtgui-${PV}[gles2=]
+	~dev-qt/qtnetwork-${PV}
+"
+RDEPEND="${DEPEND}"
+
+src_configure() {
+	local myconf=(
```
Pesa: `myconf` is only meaningful in qtbase modules. The `[gles2=]` USE dep should be enough.

dev-qt/qtdatavis3d/qtdatavis3d-5.7.0.ebuild
```diff
@@ -0,0 +1,29 @@
...
+	KEYWORDS="~amd64 ~x86"
+fi
+
+IUSE="gles2"
+
+DEPEND="
+	~dev-qt/qtcore-${PV}
+	~dev-qt/qtdeclarative-${PV}[gles2=]
+	~dev-qt/qtgui-${PV}[gles2=]
+	~dev-qt/qtnetwork-${PV}
```
Pesa: I don't see any references to qtnetwork in the code...

Pesa:
There is a hard qtwidgets dependency in `qtdatavis3d.pro`, but it doesn't actually seem to be used outside examples and tests...

Pesa:
Also reformat the commit msg

Pesa:
You haven't addressed the qtwidgets dep issue

tbrethom:
Done

Pesa:
Sorry, that's not what I meant.

There's a `requires(qtHaveModule(widgets))` at the beginning of `qtdatavis3d.pro`. That test will fail if qtwidgets is not installed. However, since qtwidgets doesn't seem to be actually used, we could sed that line one, instead of adding a dep that would be unused.

Pesa:
Amended and merged, thanks! 2fc29bba0c357762a108f9bfe390d7470245b4c7

PR #129, opened 2016-07-07
[dev-qt/qtvirtualkeyboard] New package.

tbrethom:


dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+LANGS="ar da de en es fa fi fr hi it ja ko nb pl pt ru sv zh_CN zh_TW"
+
+DESCRIPTION="Virtual keyboard plugin for the Qt5 framework"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~x86"
+fi
+
+IUSE="+hunspell"
```
Pesa: probably just "spell"... hunspell/aspell should be used only to distinguish among different spell-checking providers when the package supports more than one.

dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+DEPEND="
+	hunspell? ( app-text/hunspell )
```
Pesa: move this last

Pesa:
Also reformat the commit msg

dev-qt/qtvirtualkeyboard/metadata.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="project">
+		<email>qt@gentoo.org</email>
+		<name>Gentoo Qt Project</name>
+	</maintainer>
+	<use>
+		<flag name="spell">Adds dictionary integration support</flag>
```
Pesa: no need to describe it here, it's a global USE flag

dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+RDEPEND="${DEPEND}"
+
+for X in ${LANGS} ; do
+	IUSE="${IUSE} linguas_${X}"
+done
+
+src_configure() {
+
+	local myqmakeargs=(
+		$(usex spell '' 'CONFIG+=disable-hunspell')
```
Pesa: careful because it seems that `virtualkeyboard.pro` prefers the bundled copy of hunspell

dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=6
+inherit qt5-build
+
+LANGS="ar da de en es fa fi fr hi it ja ko nb pl pt ru sv zh_CN zh_TW"
```
Pesa: locales should be handled via `l10n.eclass`

dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	KEYWORDS="~amd64 ~x86"
+fi
+
+IUSE="+spell"
+
+DEPEND="
+	~dev-qt/qtcore-${PV}
+	~dev-qt/qtdeclarative-${PV}
+	~dev-qt/qtgui-${PV}
+	~dev-qt/qtnetwork-${PV}
```
Pesa: I don't see this dependency anywhere in the sources, but please correct me if I'm wrong, I just took a very quick look

dev-qt/qtvirtualkeyboard/qtvirtualkeyboard-5.7.0.ebuild
```diff
@@ -0,0 +1,43 @@
...
+	spell? ( app-text/hunspell )
+"
+
```
Pesa: delete this blank line

Pesa:
This requires quite a bit more work... there's a ton of stuff in `src/virtualkeyboard/3rdparty` that needs to be checked for bundled libraries.

I see at least two interesting CONFIG flags: `handwriting` (not critical, can be implemented later), and `disable-desktop` (seems to get rid of a direct xcb dep?). Plus there's a `arrow-key-navigation` option, I'm not sure what it does and if we care about it...

Pesa:
Also, are the styles (`src/virtualkeyboard/styles/styles.pro`) built by default? If not, we should probably build them.

tbrethom:
Here are some informations about the various build options : 
http://doc.qt.io/qt-5/build.html#build-instructions

The language part does'nt build locales but add some input methods for soman langauges (zh_TW for instance).

The hunspell support uses the system library if available, so i don't think we need to worry there...

tbrethom:
I've added the handwriting and xcb use flags as requested, but i'll need some advice about how to handle the 3rdparty libs properly.

a17r:
https://bugreports.qt.io/browse/QTBUG-59594

a17r:
@tbrethom ping - are you still looking into this packaging effort?

a17r:
I did look at what opensuse are doing, and they only select languages that do not force-enable 3rdparty libs:

```
# Only enable languages that don't force-include bundled 3rdparty libs
%qmake5 "CONFIG+=lang-ar_AR lang-da_DK lang-de_DE lang-en_GB lang-es_ES lang-fa_FA lang-fi_FI lang-fr_FR lang-hi_IN lang-it_IT lang-nb_NO lang-pl_PL lang-pt_PT lang-ro_RO lang-ru_RU lang-sv_SE"
```

https://build.opensuse.org/package/view_file/openSUSE:Factory/libqt5-qtvirtualkeyboard/libqt5-qtvirtualkeyboard.spec?expand=1

a17r:
Sorry, I didn't realise the leftover line in my personal overlay (neither did I know it would close this PR).

Pesa:
Superseded by https://github.com/gentoo/qt/pull/155

PR #130, opened 2016-07-25
dev-qt/qtnetwork : Libressl support for 5.7

waltercool:
Dirty but working support for LibreSSL on QtNetwork 5.7.

### DIFF
--- a/dev-qt/qtnetwork/files/qtnetwork-5.7-libressl.patch
+++ b/dev-qt/qtnetwork/files/qtnetwork-5.7-libressl.patch
@@ -1,0 +1,53 @@
+From 81494e67eccba04fc3fe554d76a9ca6fe7f2250e Mon Sep 17 00:00:00 2001
+From: hasufell <hasufell@gentoo.org>
+Date: Sat, 10 Oct 2015 01:15:01 +0200
+Subject: [PATCH] Fix compilation with libressl
+
+By additionally checking for defined(SSL_CTRL_SET_CURVES), which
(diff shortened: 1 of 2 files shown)


Pesa:
I would be inclined to reject this, see discussion in https://bugs.gentoo.org/show_bug.cgi?id=562050

@kensington are you still of the same opinion?

waltercool:
@Pesa This patch is just intended for 5.7, not for 5.9999 or 5.7.9999 because the same reason is explained in Gentoo bugzilla, QtNetworking changes with each version, and the patch needs to be different each time. I been doing a specific patch in my local overlay for 5.5, 5.6 and now for 5.7.

Until now I haven't seen any problem with QtWebBrowser, KDE Frameworks, or other KDE Application.

kensington:
> I would be inclined to reject this, see discussion in https://bugs.gentoo.org/show_bug.cgi?id=562050
> 
> @kensington are you still of the same opinion?

Yep, upstream's position is reasonable and I don't think we should hack around that.

Pesa:
Closing then.

Pesa:
> QtNetworking changes with each version, and the patch needs to be different each time. I been doing a specific patch in my local overlay for 5.5, 5.6 and now for 5.7.

I'm aware of that, and it's yet another reason why we (I) don't want to carry a downstream patch.

Moreover this is very very delicate crypto stuff... I don't want to be the next distro maintainer who introduced a security bug into a package because he patched some code he didn't fully understand. Sorry.

PR #132, opened 2016-10-06
dev-qt/qtcore: Fix installing 5.6.9999 and 5.7.9999.

Chiitoo:
Fixes install-phase that was broken due to the following upstream change:

http://code.qt.io/cgit/qt/qtbase.git/commit/?id=a7ddef139415f74f9ba8dc84a2f15105149ca5e8

Gentoo-Bug: https://bugs.gentoo.org/596054

Have not tested 5.6.9999 outside of an 'ebuild install' run, which went fine, and I really don't know if this is a good fix to put in place or not.

I took a stab at it, and since it seems to work OK here so far, I decided to share.  :]

### DIFF
--- a/dev-qt/qtcore/qtcore-5.6.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.6.9999.ebuild
@@ -25,6 +25,7 @@ DEPEND="
 RDEPEND="${DEPEND}"
 
 QT5_TARGET_SUBDIRS=(
+	doc
 	src/tools/bootstrap
 	src/tools/moc
 	src/tools/rcc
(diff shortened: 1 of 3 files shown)


dev-qt/qtcore/qtcore-5.6.9999.ebuild
```diff
@@ -25,6 +25,7 @@ DEPEND="
 RDEPEND="${DEPEND}"
 
 QT5_TARGET_SUBDIRS=(
+	doc
```
Pesa: any reason why you added this as first target? can we put it last?

Pesa: No I wasn't waiting for that. I'm just very busy lately and haven't had the time to look into this. Sorry.

kensington:
I've tested this locally and it works for me, thanks. I've squashed in a couple of things:

* moved  ``doc`` in ``QT5_TARGET_SUBDIRS`` as per pesa's request
* reduce indent of ``else`` in eclass change by one level
* modified commit message and dropped upper bound in the eclass change as the referenced commit has since been upmerged upstream

Picked as 2253ac66c0bfa28b441242c907c03c2a8ded9ab...ff79c14ece9b5fbd2e13a897e67d37ab241a3b68.

PR #133, opened 2016-11-05
Fix compatibility with recent change in 5.7

jktjkt:
This build target got removed in qt/qtbase@a7ddef139415f74f9ba8dc84a2f15105149ca5e8 , so let's restore functionality of building this again.


kensington:
See also PR #132.

jktjkt:
OK, that one is clearly better.

PR #134, opened 2016-11-30
Introduce qt_use_disable_config

kensington:
In qt/qtbase@a668c6a6, ``!contains(QT_DISABLED_FEATURES, udpsocket) {
     SUBDIRS += tuiotouch
 } `` became ``
qtConfig(udpsocket) {
     SUBDIRS += tuiotouch
 }
``

``qtConfig`` appears to be defined such that if the feature status is undefined, it dies (``error("Could not find feature $${1}.")``).

This introduces ``qt_use_disable_config``, a clone of ``qt_use_disable_mod`` to handle this new style of automagic dependency.

### DIFF
--- a/dev-qt/qtgui/qtgui-5.8.9999.ebuild
+++ b/dev-qt/qtgui/qtgui-5.8.9999.ebuild
@@ -15,7 +15,7 @@ fi
 # TODO: linuxfb
 
 IUSE="accessibility dbus egl eglfs evdev +gif gles2 gtk
-	ibus jpeg libinput +png tslib tuio +udev +xcb"
+	ibus jpeg libinput +png tslib tuio +udev vnc +xcb"
 REQUIRED_USE="
 	|| ( eglfs xcb )
 	accessibility? ( dbus xcb )
@@ -56,6 +56,7 @@ RDEPEND="
 	tslib? ( x11-libs/tslib )
 	tuio? ( ~dev-qt/qtnetwork-${PV} )
 	udev? ( virtual/libudev:= )
+	vnc? ( ~dev-qt/qtnetwork-${PV} )
 	xcb? (
 		x11-libs/libICE
 		x11-libs/libSM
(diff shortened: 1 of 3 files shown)


dev-qt/qtgui/qtgui-5.8.9999.ebuild
```diff
@@ -137,12 +138,14 @@ src_prepare() {
 	use dbus || sed -i -e 's/contains(QT_CONFIG, dbus)/false/' \
 		src/platformsupport/platformsupport.pro || die
 
+	qt_use_disable_config tuio udpsocket src/plugins/generic/generic.pro
+
 	qt_use_disable_mod ibus dbus \
```
Pesa: (minor thing) the "avoid automagic dep on qtdbus" comment above was supposed to refer to this line too... while the "avoid automagic dep on qtnetwork" still applies to the `tuio` stuff that you moved up...

dev-qt/qtgui/qtgui-5.8.9999.ebuild
```diff
@@ -137,12 +138,14 @@ src_prepare() {
 	use dbus || sed -i -e 's/contains(QT_CONFIG, dbus)/false/' \
 		src/platformsupport/platformsupport.pro || die
 
+	qt_use_disable_config tuio udpsocket src/plugins/generic/generic.pro
```
Pesa: there's a `QT_FOR_CONFIG += gui-private network-private` in that file. I don't know/remember what `QT_FOR_CONFIG` is, so it may not be an issue, but it needs to be investigated.

eclass/qt5-build.eclass
```diff
@@ -388,6 +388,29 @@ qt_use_compile_test() {
...
+
+	local flag=$1
+	local config=$2
+	shift 2
+
+	if ! use "${flag}"; then
+		echo "$@" | xargs sed -i -e "s/qtConfig(${config})/false/g" || die
+	fi
+}
+
```
Pesa: LGTM

dev-qt/qtgui/qtgui-5.8.9999.ebuild
kensington: I moved it because we traditionally we don't seem to have comments on function calls like this. I have no preference so I'm happy to duplicate the comment or group them together or whatever - just let me know what you prefer.

dev-qt/qtgui/qtgui-5.8.9999.ebuild
Pesa: yeah it seems it's only used to determine the set of modules that need to be searched for the required feature, so we're good.

dev-qt/qtgui/qtgui-5.8.9999.ebuild
Pesa: yeah the comments are rather obvious so feel free to just ditch them

Pesa: and go ahead and merge it once you've done that. Thanks!

kensington:
Landed as ff79c14ece9b5fbd2e13a897e67d37ab241a3b68...3c7ccc40ac224ac086c1660342eb6f06669a402a.

PR #135, opened 2016-12-01
dev-qt/qtgui: move gtk platform plugin to qtwidgets

kensington:
src/plugins/platformthemes/platformthemes.pro now looks like this:

```TEMPLATE = subdirs
QT_FOR_CONFIG += widgets-private

qtHaveModule(widgets):qtConfig(gtk3): SUBDIRS += gtk3
```

### DIFF
--- a/dev-qt/qtwidgets/qtwidgets-5.8.9999.ebuild
+++ b/dev-qt/qtwidgets/qtwidgets-5.8.9999.ebuild
@@ -13,26 +13,35 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
 fi
 
 # keep IUSE defaults in sync with qtgui
-IUSE="gles2 +png +xcb"
+IUSE="gles2 gtk +png +xcb"
 
 DEPEND="
 	~dev-qt/qtcore-${PV}
 	~dev-qt/qtgui-${PV}[gles2=,png=,xcb?]
+	gtk? (
+		x11-libs/gtk+:3
+		x11-libs/libX11
+		x11-libs/pango
+		!!x11-libs/cairo[qt4]
+	)
 "
 RDEPEND="${DEPEND}"
 
 QT5_TARGET_SUBDIRS=(
 	src/tools/uic
 	src/widgets
+	src/plugins/platformthemes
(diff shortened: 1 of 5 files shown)


Pesa:
Well technically the plugin still doesn't depend on qtwidgets, but it does make sense to have it in the qtwidgets package... maybe mention qtbase commit 2d3c73fcfe7a93cb46190e8e82410fe93145dbe0 in the commit msg

Pesa:
There's a specialized description for the USE flag in `metadata.xml` that you should copy over.

dev-qt/qtwidgets/qtwidgets-5.8.9999.ebuild
```diff
@@ -13,26 +13,35 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
...
+IUSE="gles2 gtk +png +xcb"
 
 DEPEND="
 	~dev-qt/qtcore-${PV}
 	~dev-qt/qtgui-${PV}[gles2=,png=,xcb?]
+	gtk? (
+		x11-libs/gtk+:3
+		x11-libs/libX11
+		x11-libs/pango
+		!!x11-libs/cairo[qt4]
```
Pesa: do we have to keep this blocker? how long ago was USE=qt4 dropped from cairo?

kensington:
I did mention the change id in the message, will update with commit hash if you prefer. metadata.xml updated. Looks like cairo[qt4] disappeared in gentoo/gentoo@0064e9d163f4458b17981d3a1efa7e004d387964 in August 2015.

Pesa:
> I did mention the change id in the message, will update with commit hash if you prefer.

yeah I'd rather have the commit sha than the change id

> Looks like cairo[qt4] disappeared in gentoo/gentoo@0064e9d in August 2015

and IIRC the useflag was masked for a long time before that, so I'd say it's safe to drop the blocker

Pesa:
Feel free to merge btw

kensington:
Thanks, picked as 09ec279994a1c99582b923c98deb9e6bafd551e6.

PR #136, opened 2016-12-01
qt5-build.eclass: fix avoiding qmake recompilation for Qt 5.8 and later

kensington:
The old hack is no longer possible since qt/qtbase@4ce0beee1b69a8695fc24a244a8a3053711906ac. This is a much worse hack but I'm not sure what other option we have.

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -179,8 +179,14 @@ qt5-build_src_prepare() {
 		qt5_symlink_tools_to_build_dir
 
 		# Avoid unnecessary qmake recompilations
-		sed -i -re "s|^if true;.*(\[ '\!').*(\"\\\$outpath/bin/qmake\".*)|if \1 -e \2 then|" \
-			configure || die "sed failed (skip qmake bootstrap)"
+		if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
+			sed -i -e "/Creating qmake/i if [ '!' -e \"\$outpath/bin/qmake\" ]; then" \
+				-e '/echo "Done."/a fi' \
+				configure || die "sed failed (skip qmake bootstrap)"
+		else
+			sed -i -re "s|^if true;.*(\[ '\!').*(\"\\\$outpath/bin/qmake\".*)|if \1 -e \2 then|" \
+				configure || die "sed failed (skip qmake bootstrap)"
+		fi
 
 		# Respect CC, CXX, *FLAGS, MAKEOPTS and EXTRA_EMAKE when bootstrapping qmake
 		sed -i -e "/outpath\/qmake\".*\"\$MAKE\")/ s:): \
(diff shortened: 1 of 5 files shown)


eclass/qt5-build.eclass
```diff
@@ -183,8 +183,13 @@ qt5-build_src_prepare() {
 		qt5_symlink_tools_to_build_dir
 
 		# Avoid unnecessary qmake recompilations
-		sed -i -re "s|^if true;.*(\[ '\!').*(\"\\\$outpath/bin/qmake\".*)|if \1 -e \2 then|" \
-			configure || die "sed failed (skip qmake bootstrap)"
+		if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
+			sed -e "/Creating qmake/i if false; then" -e '/cd "$outpath\/qmake"; "$MAKE" -s/a fi' \
```
Pesa: why not adding `fi` after `echo "Done."`?

also, keep one sed expression per line, so break before the second `-e`, and use single quotes for the first sed expression as well

eclass/qt5-build.eclass
```diff
@@ -183,8 +183,13 @@ qt5-build_src_prepare() {
...
+				-i configure || die "sed failed (skip qmake bootstrap)"
```
Pesa: put the `-i` right after `sed` on the previous line, for consistency

kensington:
Updated.

eclass/qt5-build.eclass
```diff
@@ -179,8 +179,14 @@ qt5-build_src_prepare() {
 		qt5_symlink_tools_to_build_dir
 
 		# Avoid unnecessary qmake recompilations
-		sed -i -re "s|^if true;.*(\[ '\!').*(\"\\\$outpath/bin/qmake\".*)|if \1 -e \2 then|" \
-			configure || die "sed failed (skip qmake bootstrap)"
+		if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
+			sed -i -e '/Creating qmake/i if false; then' \
```
Pesa: hmm this disables the build unconditionally though right? how does qmake get built in `qtcore`?

kensington: Forgive me, I completely missed the purpose of the original code.

eclass/qt5-build.eclass
```diff
@@ -179,8 +179,14 @@ qt5-build_src_prepare() {
 		qt5_symlink_tools_to_build_dir
 
 		# Avoid unnecessary qmake recompilations
-		sed -i -re "s|^if true;.*(\[ '\!').*(\"\\\$outpath/bin/qmake\".*)|if \1 -e \2 then|" \
-			configure || die "sed failed (skip qmake bootstrap)"
+		if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
+			sed -i -e '/Creating qmake/i if [ '!' -e "$outpath/bin/qmake" ]; then' \
```
Pesa: hmm there's a quoting problem here, I'm afraid you have to use double quotes as outermost quotes (for the whole sed expression) and escape them when used as literal quotes around `$outpath/bin/qmake`. Also don't forget to escape the `$`.

kensington: Indeed, I'll fix in a minute. Curiously, it does still work however!

kensington:
Picked as 241a0712deb4e10edaa70f6681490a97d49d5126.

PR #137, opened 2016-12-06
dev-qt/qtcore: disable iconv if icu is enabled

kensington:
See qt/qtbase@fd9e5d90333c385ad191a289b2e95e918c58b242.

I wonder if we should stop passing iconv in the eclass, then.

### DIFF
--- a/dev-qt/qtcore/qtcore-5.8.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.8.9999.ebuild
@@ -19,8 +19,8 @@ DEPEND="
 	dev-libs/glib:2
 	>=dev-libs/libpcre-8.38[pcre16,unicode]
 	>=sys-libs/zlib-1.2.5
-	virtual/libiconv
 	icu? ( dev-libs/icu:= )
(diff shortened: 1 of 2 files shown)


dev-qt/qtcore/qtcore-5.8.9999.ebuild
```diff
@@ -36,6 +36,7 @@ QT5_TARGET_SUBDIRS=(
 
 src_configure() {
 	local myconf=(
+		$(usex icu '-no-iconv' '-iconv')
```
Pesa: no need to quote

Pesa: we could also just merge this with the following `qt_use icu` in a single `usex icu ...`

Pesa:
> I wonder if we should stop passing iconv in the eclass, then.

I guess so, conditionally as usual

kensington:
Fixed in fd78df39dfe6fbc13d8e3358157dcf3201ef12bc.

PR #138, opened 2016-12-14
www-client/qupzilla: Bump Qt dep to 5.8 and update PLOCALES.

Chiitoo:
Qt 5.7 is required since an upstream change made on 2016-10-24: https://github.com/QupZilla/qupzilla/commit/cf3fa88942926a0db72164ed1bcd1d2d4b59c125

### DIFF
--- a/www-client/qupzilla/qupzilla-9999.ebuild
+++ b/www-client/qupzilla/qupzilla-9999.ebuild
@@ -5,10 +5,10 @@
 EAPI=6
 
 PLOCALES="ar_SA bg_BG ca_ES cs_CZ da_DK de_DE el_GR es_ES es_MX es_VE eu_ES
-	fa_IR fi_FI fr_FR gl_ES he_IL hr_HR hu_HU id_ID it_IT ja_JP ka_GE lg
-	lt lv_LV nl_NL nqo pl_PL pt_BR pt_PT ro_RO ru_RU sk_SK sr
+	fa_IR fi_FI fr_FR gl_ES he_IL hr_HR hu_HU id_ID is it_IT ja_JP ka_GE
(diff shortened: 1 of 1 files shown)


Chiitoo:
The requirement is now 5.8, as of 13cba7a069f208dd1b03b3426911a40f882d7bf2.

I don't imagine there is point in updating (or pushing) this request until a time that we finally figure out at least 'dev-qt/dbus'...

Pesa:
If that's the minimum requirement, we should still update the dependency. The fact that `qtdbus-5.8.0` is broken is somewhat orthogonal...
I don't like how quickly upstream is dropping support for "older" versions of Qt, but that's a separate issue.

Chiitoo:
Yeah.  I wonder if it could be due to using features only in QtWebEngine 5.8, with unwillingness of keeping things optional depending on the used version.  I'll ask them when I see them around.

Should I squash the commits or keep the history, or open a whole new request with the 5.8 (and PLOCALES) updates?

Pesa:
squash

Chiitoo:
Confirmed the upstream reasoning to indeed be that it would simply be much work to keep older versions of QtWebEngine supported.

Updated commits.

Remembered we actually have a 'dbus' USE-flag to make 'qtdbus' optional here...

Pesa:
You should also consider taking (co-)maintainership of this package :)

Chiitoo:
I'm a staffer, yo!

Don't have git access etc.

kensington:
Thanks, picked as 0339b82dc658d36fe2a50dde050e11d530a075bd...4b10dabef17239fe9c818d57e3fe4e02c4ce35b4.

PR #140, opened 2017-03-01
dev-qt/qtdbus: Adjust the awful hack to make 'qtdbus' build.

Chiitoo:
For the live-builds, the files we want to remove will be under the build directory, not the source directory.  Without this change, the files are not found.

For at least 5.8.0, it seems like it might be enough to have the 'src/corelib' added to 'QT5_TARGET_SUBDIRS'.  I'm not super sure of that yet, so I may be missing something more or/and less obvious.

### DIFF
--- a/dev-qt/qtdbus/qtdbus-5.8.9999.ebuild
+++ b/dev-qt/qtdbus/qtdbus-5.8.9999.ebuild
@@ -53,7 +53,7 @@ src_compile() {
 	hack() {
 		emake
 		if [[ ${subdir} = "src/corelib" ]]; then
-			rm "${S}"/lib/libQt5Core* || die
+			rm "${QT5_BUILD_DIR}"/lib/libQt5Core* || die
 		fi
 	}
 	qt5_foreach_target_subdir hack
--- a/dev-qt/qtdbus/qtdbus-5.9999.ebuild
+++ b/dev-qt/qtdbus/qtdbus-5.9999.ebuild
@@ -53,7 +53,7 @@ src_compile() {
 	hack() {
 		emake
 		if [[ ${subdir} = "src/corelib" ]]; then
-			rm "${S}"/lib/libQt5Core* || die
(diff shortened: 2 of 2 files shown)


dev-qt/qtdbus/qtdbus-5.8.9999.ebuild
```diff
@@ -53,7 +53,7 @@ src_compile() {
 	hack() {
 		emake
 		if [[ ${subdir} = "src/corelib" ]]; then
-			rm "${S}"/lib/libQt5Core* || die
+			rm "${WORKDIR}/${P}_build"/lib/libQt5Core* || die
```
Pesa: use `${QT5_BUILD_DIR}`

dev-qt/qtdbus/qtdbus-5.9999.ebuild
```diff
@@ -53,7 +53,7 @@ src_compile() {
 	hack() {
 		emake
 		if [[ ${subdir} = "src/corelib" ]]; then
-			rm "${S}"/lib/libQt5Core* || die
+			rm "${WORKDIR}/${P}_build"/lib/libQt5Core* || die
```
Pesa: ditto

Chiitoo:
Thanks.  Changes done.

Should have figured that one out.

kensington:
Thanks, merged.

kensington:
> For at least 5.8.0, it seems like it might be enough to have the 'src/corelib' added to 'QT5_TARGET_SUBDIRS'. I'm not super sure of that yet, so I may be missing something more or/and less obvious.

While it should build fine there will almost certainly be file collisions with ``dev-qt/qtcore``.

Chiitoo:
> While it should build fine there will almost certainly be file collisions with dev-qt/qtcore.

Ah, right.  Did not think of that, at all.  :]

Thanks!

PR #141, opened 2017-03-01
Fix configure for {5,5.8}.9999 'qtbase' packages.

Chiitoo:
Upstream has removed an “unused Xlib's XRender dependency”, which results into 'qtcore-5.8.9999' and 'qtgui-5.8.9999' configure phases failing due to an unknown command line option (-xrender or -no-xrender).

http://code.qt.io/cgit/qt/qtbase.git/commit/?id=d37c353dc0f2ae5bb803fe9e5752eff846246439

### DIFF
--- a/dev-qt/qtgui/qtgui-5.8.9999.ebuild
+++ b/dev-qt/qtgui/qtgui-5.8.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
@@ -55,7 +55,6 @@ RDEPEND="
 		x11-libs/libSM
 		x11-libs/libX11
 		>=x11-libs/libXi-1.7.4
-		x11-libs/libXrender
 		>=x11-libs/libxcb-1.10:=[xkb]
 		>=x11-libs/libxkbcommon-0.4.1[X]
 		x11-libs/xcb-util-image
@@ -118,7 +117,6 @@ QT5_GENTOO_CONFIG=(
 	xcb:xcb-xlib:
 	xcb:xinput2:
 	xcb::XKB
-	xcb:xrender
 )
 
 src_prepare() {
@@ -160,7 +158,7 @@ src_configure() {
 		$(qt_use udev libudev)
 		$(qt_use xcb xcb system)
 		$(qt_use xcb xkbcommon-x11 system)
-		$(usex xcb '-xcb-xlib -xinput2 -xkb -xrender' '')
+		$(usex xcb '-xcb-xlib -xinput2 -xkb' '')
 	)
 	qt5-build_src_configure
 }
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -654,7 +654,9 @@ qt5_base_configure() {
 
 		# disable undocumented X11-related flags, override in qtgui
 		# (not shown in ./configure -help output)
-		-no-xkb -no-xrender
+		-no-xkb
+		$([[ ${QT5_MINOR_VERSION} -lt 8 || ${QT5_MINOR_VERSION} -eq 8
+			&& ${QT5_PATCH_VERSION} -lt 1 ]] && echo -no-xrender)
 
 		# disable obsolete/unused X11-related flags
 		$([[ ${QT5_MINOR_VERSION} -lt 8 ]] && echo -no-mitshm -no-xcursor -no-xfixes -no-xrandr -no-xshape -no-xsync)
--- a/dev-qt/qtgui/qtgui-5.9999.ebuild
+++ b/dev-qt/qtgui/qtgui-5.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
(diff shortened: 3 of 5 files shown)


eclass/qt5-build.eclass
```diff
@@ -655,7 +655,8 @@ qt5_base_configure() {
 
 		# disable undocumented X11-related flags, override in qtgui
 		# (not shown in ./configure -help output)
-		-no-xkb -no-xrender
+		$([[ ${PV} != 5.8.9999 ]] && echo -no-xrender)
```
Pesa: This test is wrong. See how `QT5_MINOR_VERSION` and `QT5_PATCH_VERSION` are used elsewhere in the eclass.

Moreover, while it is true that that upstream commit is only present in the 5.8 branch right now, it will soon be merged into 5.9 and dev (5.9999), so you should check for `minor < 8 || (minor == 8 && patch < 1)`.

dev-qt/qtgui/qtgui-5.8.9999.ebuild
```diff
@@ -161,7 +161,7 @@ src_configure() {
 		$(qt_use udev libudev)
 		$(qt_use xcb xcb system)
 		$(qt_use xcb xkbcommon-x11 system)
-		$(usex xcb '-xcb-xlib -xinput2 -xkb -xrender' '')
+		$(usex xcb '-xcb-xlib -xinput2 -xkb' '')
```
Pesa: apply the same change to 5.9999

eclass/qt5-build.eclass
Chiitoo: Yeah, I was almost sure there was a better way to do it, but couldn't spot it yet.

And indeed, it will soon be needed for the other versions too, so I wasn't entirely sure how much sense it makes to do this now.

Pesa: It's a live ebuild, it's expected to be broken from time to time :)

Chiitoo:
Changes done.

Does this look better or worse now. I wonder.  :]

Chiitoo:
Added a commit to remove '-xrender' from 'qtwidgets' as well.

I didn't think of removing the '# $Id$' earlier... but I imagine someone will get rid of those in bulk at some point in time (or should I be taking them off while I poke at things?).

Pesa:
Does this mean that we can remove the `x11-libs/libXrender` dependency from qtgui?

eclass/qt5-build.eclass
```diff
@@ -655,7 +655,9 @@ qt5_base_configure() {
 
 		# disable undocumented X11-related flags, override in qtgui
 		# (not shown in ./configure -help output)
-		-no-xkb -no-xrender
+		-no-xkb
+		$([[ ${QT5_MINOR_VERSION} -lt 8 ]] || $([[ ${QT5_MINOR_VERSION} == 8
```
Pesa: * why the nested `$(...)`? can't you do everything inside a single `[[ ]]`? (with parentheses if needed)
* use `-eq` instead of `==` for consistency
* change the commit msg, these options are used for *all* qtbase packages, not just {qtcore,qtgui}

Pesa:
What about the `xrender` in QT5_GENTOO_CONFIG in qtgui?

eclass/qt5-build.eclass
Chiitoo: - No idea...  It was probably the first thing that worked...  Adjusting to just one test (still seem to need the one '$(...)', or things seem to break, but that's probably to be expected).
- That was a silly mistake!  Using '-eq' now.
- Good point!  Changing.

Chiitoo:
Tested building 'qtgui' without 'x11-libs/libXrender' and 'xcb:xrender', which resulted into no issues that I could spot.

All changes should be done.

Pesa:
> Tested building 'qtgui' without 'x11-libs/libXrender' and 'xcb:xrender', which resulted into no issues that I could spot.

Those issues tend to be hard to find. The fact that it builds isn't enough. Can you also check the following please?

1. `-lXrender` is not passed to any linker command lines during build
2. run `scanelf -nq` on all shared objects built by dev-qt/qtgui (`libQt5XcbQpa.so` should be the interesting one) and verify that `libXrender.so` does not appear in the output

Chiitoo:
I figured there were more ways I could/should have checked things up.  Thanks!

So far it looks like good riddance:

- https://gist.github.com/Chiitoo/b2b836d2ff52b5c602d2e67666d2ed1b
- https://gist.github.com/Chiitoo/5dc485732f24f9d1a693f5f2f6393c6a

That's for '5.8.9999'.  I did not check with '5.9999' yet, because configure fails at the 'libs.pcre2' test though I have it installed...  Gotta look more into that.

Pesa:
> https://gist.github.com/Chiitoo/b2b836d2ff52b5c602d2e67666d2ed1b
> https://gist.github.com/Chiitoo/5dc485732f24f9d1a693f5f2f6393c6a

Great, nothing unexpected here.

>  I did not check with '5.9999' yet

Don't worry about 5.9999, I'm sure they didn't re-add the dependency there ;)

kenya888:
Hi, what prevents this PR from merging?

Pesa:
> Hi, what prevents this PR from merging?

Probably nothing, apart from the fact that I've been traveling... I'll take another look at this in the next few days.

kenya888:
@Pesa Great. Have a nice trip :smile:

Pesa:
Finally merged! Thanks and sorry for the delay!

Chiitoo:
\ö/

Sorry on my part for the obvious things that may have made it take longer than necessary.  :]

PR #142, opened 2017-03-13
PySide2: From the Dead We Shall Resurrect You

leycec:
This pull request fixes all issues reported at [bug 612272](https://bugs.gentoo.org/show_bug.cgi?id=612272) on [Bugzilla](https://bugs.gentoo.org) *and* several other minor issues since uncovered. <sup>*details below*</sup>

Before we begin, I'd like to repeat the opening refrain of that bug report: **thank you.** This overlay is a diamond in the cruft, whose tireless volunteers have made all things possible again. Thanks, all!

## Just Tell Us What's Wrong Already

The current state of the PySide2 ebuild is… *less than desirable.* It's an excellent foundation that currently doesn't work at all. Let's see if we can do something about that.

Outstanding issues include (*in descending order of importance*):

* **Package namespace collision.** This overlay's Qt5-specific `dev-python/pyside` and `dev-python/shiboken` packages collide with Portage's Qt4-specific packages of the same name, preventing PySide1 and PySide2 from being concurrently installed *and* breaking downstream ebuilds requiring only PySide1. PySide1 and PySide2 are similar but ultimately incompatible (much like Python 2 and 3). Porting between the two is feasible but non-trivial. Existing ebuilds requiring `dev-python/pyside` typically expect PySide1 and hence are *not* satisfied by this overlay's `dev-python/pyside-9999` ebuild.
* **All USE flags are currently ignored.** <sup>_It hurts._</sup>
* The new optional `Qt5WebChannel`, `Qt5WebEngineWidgets`, and `Qt5WebSockets` dependencies have no corresponding USE flags.
* The `${QT_PV}` global implies PySide2 to only require Qt >= 4.8.5, when in fact PySide2 requires Qt >= 5.3.0.
* The `${DOCS}` global references a `ChangeLog` file that no longer exists.
* The optional `dev-qt/qtsvg:5` dependency no longer accepts an `accessibility` USE flag.
* The `${PYTHON_COMPAT}` globals incorrectly omit Python 3.6, which is [explicitly supported](https://bugreports.qt.io/browse/PYSIDE-471).
* Ebuild metadata (e.g., `${DESCRIPTION}`, `${HOMEPAGE}`) incorrectly references PySide1.

See the [bug report](https://bugs.gentoo.org/show_bug.cgi?id=612272) for tedious details so boring you'll want to [punch a baby](https://www.youtube.com/watch?v=otn96q_v6P0).

## Just Tell Us What You Did Already

This pull request addresses *most* of the above issues – excluding the new optional dependencies with no USE flags. I just ignored those. I'm lazy and like video games.

Specifically (*in the same order as above*):

* This overlay's Qt5-specific `dev-python/pyside` and `dev-python/shiboken` packages have been moved to `dev-python/pyside2` and `dev-python/shiboken2`, coinciding with upstream path, package, and project names. To facilitate this move, file collisions with users installing the older `dev-python/pyside-9999` and `dev-python/shiboken-9999` ebuilds have been prevented as follows:
  * `dev-python/pyside2-9999` is blocked by `dev-python/pyside-9999`.
  * `dev-python/shiboken2-9999` is blocked by `dev-python/shiboken-9999`.
* All USE flags are now handled as expected. This was probably the least trivial change. In the `src_configure()` phase:
  * Each `cmake` option of the form `-DDISABLE_${DEPENDENCY}=$(usex !${USE_FLAG})` has been converted into an equivalent option of the form `-DCMAKE_DISABLE_FIND_PACKAGE_${DEPENDENCY}=$(usex !X)`. For details on the magic `CMAKE_DISABLE_FIND_PACKAGE_${DEPENDENCY}` global, see CMake's [official documentation](https://cmake.org/cmake/help/v3.0/command/find_package.html) for the `find_package()` function.
  * Each Qt4-specific dependency has been converted into the equivalent Qt5-specific dependency.
  * Several dependencies optional under PySide1 are now mandatory under PySide2, including `Qt5Gui`, `Qt5Sql`, and `Qt5Test`. Yes, that's right: **PySide2 requires both X and SQL support,** at least superficially. I'm inclined to believe that this is an upstream bug. PyQt5 does *not* require SQL and there's little reason to believe that PySide2 should either. This should probably be reported to the [upstream issue tracker](https://bugreports.qt.io/browse/PYSIDE/fixforversion/15785). Until resolved upstream, the PySide2 ebuild now patches the `PySide2/CMakeLists.txt` file to only optionally require these dependencies.
* The `${QT_PV}` global now requires Qt >= 5.3.0.
* The `${DOCS}` global has been deleted.
* The optional `dev-qt/qtsvg:5` dependency is now passed no USE flags.
* Python 3.6 is now supported by both the PySide2 and Shiboken2 ebuilds.
* Ebuild metadata (e.g., `${DESCRIPTION}`, `${HOMEPAGE}`) has been updated.
* Copyright headers have been updated to 2017.

And that's how the cookie was crumbled. :cookie:

### DIFF
--- a/dev-python/pyside/metadata.xml
+++ b/dev-python/pyside/metadata.xml
@@ -6,18 +6,26 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
-		<flag name="X">Build QtGui and QtTest modules</flag>
-		<flag name="declarative">Build QtDeclarative module</flag>
+		<flag name="concurrent">Build QtConcurrent module</flag>
+		<flag name="declarative">Build QtQml, QtQuick, and QtQuickWidgets modules</flag>
 		<flag name="designer">Build QtDesigner and QtUiTools modules</flag>
+		<flag name="gui">Build QtGui module</flag>
 		<flag name="help">Build QtHelp module</flag>
 		<flag name="multimedia">Build QtMultimedia module</flag>
+		<flag name="network">Build QtNetwork module</flag>
 		<flag name="opengl">Build QtOpenGL module</flag>
+		<flag name="printsupport">Build QtPrintSupport module</flag>
 		<flag name="script">Build QtScript module</flag>
 		<flag name="scripttools">Build QtScriptTools module</flag>
 		<flag name="sql">Build QtSql module</flag>
 		<flag name="svg">Build QtSvg module</flag>
-		<flag name="webkit">Build QtWebKit module</flag>
+		<flag name="testlib">Build QtTest module</flag>
+		<flag name="webchannel">Build QtWebChannel module</flag>
+		<flag name="webengine">Build QtWebEngine and QtWebEngineWidgets modules</flag>
+		<flag name="webkit">Build QtWebKit and QtWebKitWidgets modules</flag>
+		<flag name="websockets">Build QtWebSockets module</flag>
+		<flag name="widgets">Build QtWidgets module</flag>
+		<flag name="x11extras">Build QtX11Extras module</flag>
 		<flag name="xmlpatterns">Build QtXmlPatterns module</flag>
 	</use>
 </pkgmetadata>
-
--- a/dev-python/pyside/pyside-9999.ebuild
+++ b/dev-python/pyside/pyside-9999.ebuild
@@ -1,71 +1,97 @@
-# Copyright 1999-2016 Gentoo Foundation
+# Copyright 1999-2017 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
 
-PYTHON_COMPAT=( python{2_7,3_4,3_5} )
+PYTHON_COMPAT=( python{2_7,3_4,3_5,3_6} )
 
 inherit cmake-utils flag-o-matic python-r1 virtualx git-r3
 
 DESCRIPTION="Python bindings for the Qt framework"
-HOMEPAGE="https://wiki.qt.io/Pyside"
+HOMEPAGE="https://wiki.qt.io/PySide2"
 EGIT_REPO_URI=(
-	"git://code.qt.io/pyside/${PN}.git"
-	"https://code.qt.io/git/pyside/${PN}.git"
+	"git://code.qt.io/pyside/pyside.git"
+	"https://code.qt.io/git/pyside/pyside.git"
 )
 
+#FIXME: Switch to the clang-enabled "dev" branch once stable.
+EGIT_BRANCH="5.6"
+
 LICENSE="LGPL-2.1"
-SLOT="0"
+SLOT="2"
 KEYWORDS=""
 
-IUSE="X declarative designer help multimedia opengl script scripttools sql svg test webkit xmlpatterns"
+IUSE="
+	concurrent declarative designer gui help multimedia network opengl
+	printsupport script scripttools sql svg test testlib webchannel webengine
+	webkit websockets widgets x11extras xmlpatterns"
+
+# The requirements below were strongly inspired by their "PyQt5" equivalents.
 REQUIRED_USE="
 	${PYTHON_REQUIRED_USE}
-	declarative? ( X )
-	designer? ( X )
-	help? ( X )
-	multimedia? ( X )
-	opengl? ( X )
-	scripttools? ( X script )
-	sql? ( X )
-	svg? ( X )
-	test? ( X )
-	webkit? ( X )
+	declarative? ( gui network )
+	designer? ( widgets )
+	help? ( widgets )
+	multimedia? ( gui network )
+	opengl? ( widgets )
+	printsupport? ( widgets )
+	scripttools? ( gui script )
+	sql? ( widgets )
+	svg? ( widgets )
+	test? ( widgets )
+	testlib? ( widgets )
+	webchannel? ( network )
+	webengine? ( network webchannel widgets )
+	webkit? ( gui network printsupport widgets )
+	websockets? ( network )
+	widgets? ( gui )
+	xmlpatterns? ( network )
 "
 
-# Minimal supported version of Qt.
-QT_PV="4.8.5:4"
+# Minimum version of Qt required, derived from the "CMakeLists.txt" line:
+#      find_package(Qt5 ${QT_PV} REQUIRED COMPONENTS Core)
+QT_PV=":5/5.6"
 
 RDEPEND="
 	${PYTHON_DEPS}
-	>=dev-python/shiboken-${PV}[${PYTHON_USEDEP}]
-	>=dev-qt/qtcore-${QT_PV}
-	X? (
-		>=dev-qt/qtgui-${QT_PV}[accessibility]
-		>=dev-qt/qttest-${QT_PV}
-	)
-	declarative? ( >=dev-qt/qtdeclarative-${QT_PV} )
-	designer? ( >=dev-qt/designer-${QT_PV} )
-	help? ( >=dev-qt/qthelp-${QT_PV} )
-	multimedia? ( >=dev-qt/qtmultimedia-${QT_PV} )
-	opengl? ( >=dev-qt/qtopengl-${QT_PV} )
-	script? ( >=dev-qt/qtscript-${QT_PV} )
-	sql? ( >=dev-qt/qtsql-${QT_PV} )
-	svg? ( >=dev-qt/qtsvg-${QT_PV}[accessibility] )
-	webkit? ( >=dev-qt/qtwebkit-${QT_PV} )
-	xmlpatterns? ( >=dev-qt/qtxmlpatterns-${QT_PV} )
+	>=dev-python/shiboken-${PV}:${SLOT}[${PYTHON_USEDEP}]
+	dev-qt/qtcore${QT_PV}
+	dev-qt/qtxml${QT_PV}
+	declarative? ( dev-qt/qtdeclarative${QT_PV}[widgets?] )
+	designer? ( dev-qt/designer${QT_PV} )
+	help? ( dev-qt/qthelp${QT_PV} )
+	multimedia? ( dev-qt/qtmultimedia${QT_PV}[widgets?] )
+	opengl? ( dev-qt/qtopengl${QT_PV} )
+	printsupport? ( dev-qt/qtprintsupport${QT_PV} )
+	script? ( dev-qt/qtscript${QT_PV} )
+	sql? ( dev-qt/qtsql${QT_PV} )
+	svg? ( dev-qt/qtsvg${QT_PV} )
+	testlib? ( dev-qt/qttest${QT_PV} )
+	webchannel? ( dev-qt/qtwebchannel${QT_PV} )
+	webengine? ( dev-qt/qtwebengine${QT_PV}[widgets?] )
+	webkit? ( dev-qt/qtwebkit${QT_PV}[printsupport] )
+	websockets? ( dev-qt/qtwebsockets${QT_PV} )
+	x11extras? ( dev-qt/qtx11extras${QT_PV} )
+	xmlpatterns? ( dev-qt/qtxmlpatterns${QT_PV} )
+	concurrent? ( dev-qt/qtconcurrent${QT_PV} )
+	gui? ( dev-qt/qtgui${QT_PV} )
+	network? ( dev-qt/qtnetwork${QT_PV} )
+	printsupport? ( dev-qt/qtprintsupport${QT_PV} )
+	sql? ( dev-qt/qtsql${QT_PV} )
+	testlib? ( dev-qt/qttest${QT_PV} )
+	widgets? ( dev-qt/qtwidgets${QT_PV} )
 "
-DEPEND="${RDEPEND}
-	>=dev-qt/qtgui-${QT_PV}
-"
-
-DOCS=( ChangeLog )
+DEPEND="${RDEPEND}"
 
 src_prepare() {
-	# Fix generated pkgconfig file to require the shiboken
-	# library suffixed with the correct python version.
-	sed -i -e '/^Requires:/ s/shiboken$/&@SHIBOKEN_PYTHON_SUFFIX@/' \
-		libpyside/pyside2.pc.in || die
+	#FIXME: Remove the following "sed" patch after this upstream issue is closed:
+	#    https://bugreports.qt.io/browse/PYSIDE-502
+	
+	# Force the optional "Qt5Concurrent", "Qt5Gui", "Qt5Network",
+	# "Qt5PrintSupport", "Qt5Sql", "Qt5Test", and "Qt5Widgets" packages
+	# erroneously marked as mandatory to be optional.
+	sed -i -e 's/^\(CHECK_PACKAGE_FOUND(Qt5\(Concurrent\|Gui\|Network\|PrintSupport\|Sql\|Test\|Widgets\)\))$/\1 opt)/' \
+		PySide2/CMakeLists.txt || die
 
 	if use prefix; then
 		cp "${FILESDIR}"/rpath.cmake . || die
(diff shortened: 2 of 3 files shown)


leycec:
**Conflicts resolved.** Specifically, all `$Id$` header boilerplate has been dropped per 7f9090cfb43afea.

Fairwell, boilerplate. You shall not be missed.

leycec:
**Updated.** Thanks for the code review, Pesa! Unfortunately, my most recent force-push appears to have eaten up your line-by-line comments. Yikes – I had no idea code reviews were so fragile. This is what it sounds like when pull requests cry. :dove: 

On a happier note, changes in this update include:

* `dev-python/pyside2` has been renamed back to `dev-python/pyside` with `SLOT=2`. <sup>_You were [totally right.](https://bugs.gentoo.org/show_bug.cgi?id=612272)_</sup>
* `dev-python/shiboken2` has been renamed back to `dev-python/shiboken` with `SLOT=2`.
* All conditional PySide2 dependencies now have corresponding USE flags. Thanks to this, the PySide2 ebuild is now as fully-featured as the official PyQt5 ebuild. Take that, Portage!

leycec:
@Pesa Sorry for the bother. Did you need any further work from me on this?

leycec:
`<le_sigh/>`

This pull request appears to have... *stalled.* That's cool. It happens. In the interim, I've elected to **temporarily** host these ebuilds at my [personal overlay](https://github.com/leycec/raiagent).

Of course, this is non-ideal. PySide2 and Shiboken2 deserve a more officious home than my sad corner cubicle of the Internet – and this overlay is it. If this pull request is eventually merged in, I'll happily take those ebuilds down in favour of this overlay.

Until then, thanks for all the bleeding-edge Qt goodness.

dev-python/pyside/metadata.xml
```diff
@@ -6,18 +6,27 @@
...
-		<flag name="help">Build QtHelp module</flag>
-		<flag name="multimedia">Build QtMultimedia module</flag>
-		<flag name="opengl">Build QtOpenGL module</flag>
-		<flag name="script">Build QtScript module</flag>
-		<flag name="scripttools">Build QtScriptTools module</flag>
-		<flag name="sql">Build QtSql module</flag>
-		<flag name="svg">Build QtSvg module</flag>
-		<flag name="webkit">Build QtWebKit module</flag>
-		<flag name="xmlpatterns">Build QtXmlPatterns module</flag>
+		<flag name="concurrent">Build Qt5Concurrent module</flag>
```
Pesa: drop the "5" everywhere. USE flag descriptions are shared among all versions, so when pyside2 enters portage, we can't use these descriptions anymore.

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -1,71 +1,93 @@
...
+SLOT="2"
 KEYWORDS=""
 
-IUSE="X declarative designer help multimedia opengl script scripttools sql svg test webkit xmlpatterns"
+IUSE="
+	concurrent declarative designer gui help multimedia network opengl
+	printsupport script scripttools sql svg test testlib webchannel webengine
+	webkit websockets widgets x11extras xmlpatterns"
+
+# The requirements below were strongly inspired by their "PyQt5" equivalents.
```
Pesa: this is not necessarily a valid criterion, but I guess it's good enough for now...

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -1,71 +1,93 @@
...
 	# Fix generated pkgconfig file to require the shiboken
 	# library suffixed with the correct python version.
 	sed -i -e '/^Requires:/ s/shiboken$/&@SHIBOKEN_PYTHON_SUFFIX@/' \
-		libpyside/pyside2.pc.in || die
+		libpyside/${PN}2.pc.in || die
+
+	# Force the optional "Qt5Concurrent", "Qt5Gui", "Qt5Network",
+	# "Qt5PrintSupport", "Qt5Sql", "Qt5Test", and "Qt5Widgets" packages
+	# erroneously marked as mandatory to be optional.
+	sed -i -e 's/^\(CHECK_PACKAGE_FOUND(Qt5\(Concurrent\|Gui\|Network\|PrintSupport\|Sql\|Test\|Widgets\)\))$/\1 opt)/' \
```
Pesa: any chance this could be applied upstream?

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -1,71 +1,93 @@
...
-	>=dev-qt/qtgui-${QT_PV}
-"
-
-DOCS=( ChangeLog )
+DEPEND="${RDEPEND}"
 
 src_prepare() {
 	# Fix generated pkgconfig file to require the shiboken
 	# library suffixed with the correct python version.
 	sed -i -e '/^Requires:/ s/shiboken$/&@SHIBOKEN_PYTHON_SUFFIX@/' \
```
Pesa: we could/should probably drop this hack

dev-python/shiboken/shiboken-9999.ebuild
```diff
@@ -1,38 +1,41 @@
...
-SLOT="0"
+SLOT="2"
 KEYWORDS=""
 IUSE="test"
-
 REQUIRED_USE="${PYTHON_REQUIRED_USE}"
 
+# Minimum Qt version required, While PySide2 requires specific Qt 5 releases,
+# Shiboken2 purports to support all Qt 5 releases.
+QT_PV="5:5="
```
Pesa: is the subslot operator (`:=`) actually needed?

dev-python/pyside/metadata.xml
leycec: *Woopsie.* Thanks for the subtle catch. Never did like the number "5," anyway. It shall go away.

dev-python/pyside/pyside-9999.ebuild
leycec: **Heh.** Maintaining USE flag parity with `PyOt5` superficially *seemed* like a worthwhile goal, if only to reduce migraines for end users migrating from `PyQt5` to `PySide2`. No one will ever want to go in the reverse direction, of course.

That said – you're Totally Right™. `PySide2` is its own ugly beast. As time and acne accumulate, `PyQt5` and `PySide2` will probably begin diverging in non-trivial ways. For the moment, however, they're blessedly similar. <sup>_Sort of. Don't quote me on that. <sup>You're quoting me on that, aren't you?</sup>_</sup>

dev-python/pyside/pyside-9999.ebuild
leycec: I'll definitely take a deeper look at this. The `PySide2` ebuild blindly inherited this from the `PySide1` ebuild. (Maybe that's not a good thing.)

As you suggest, I'll first try dropping this kludge and see where that leaves us.

dev-python/pyside/pyside-9999.ebuild
leycec: **Absolutely.** I'm lazy, slovenly, and pressed for time at the moment. Upstream patch submission appears to be a bit of a... *process.* Cumbersome corporate bureaucracy: who would have thought?

If I fail to submit an actual patch, I'll try to at least submit a bug report suggesting these changes. :bug:

dev-python/shiboken/shiboken-9999.ebuild
leycec: Good catch. *Maybe?* I... don't know, actually.

I note that the `PySide1` ebuild omits subslot operators, but that doesn't quite seem right. Shiboken[1|2] produces the Python bindings required by PySide[1|2] by parsing the C++ headers of the desired Qt[4|5] framework. It seems to follow that updating Qt[4|5] will silently break the previously installed version of Shiboken[1|2] and hence PySide[1|2].

If true, the remedy is to automatically rebuild Shiboken[1|2] and hence PySide[1|2] on all Qt[4|5] updates. Or did I completely misunderstand subslot operators and botch this whole thing? <sup>_...not again._</sup>

dev-python/pyside/pyside-9999.ebuild
Pesa: Sorry, my comment was a little cryptic I guess, let me explain.

I do agree that maintaining USE flag parity with PyQt5 (at least as much as possible) *is* a worthwhile goal. What I meant to say here is simply that the internal inter-dependencies among pyside modules *might not* be the same as those among PyQt5 modules. But as I said, this will do for now.

dev-python/pyside/pyside-9999.ebuild
Pesa: agreed

dev-python/pyside/pyside-9999.ebuild
Pesa: yeah don't worry, we can apply this downstream for now, no rush. Maybe add a "TODO: send upstream" comment somewhere.

dev-python/shiboken/shiboken-9999.ebuild
Pesa: > It seems to follow that updating Qt[4|5] should silently break all currently installed versions of Shiboken[1|2] and hence PySide[1|2].

If shiboken only uses/wraps the public portion of the Qt API, then it should not break when updating Qt (Qt maintains backward API & ABI compatibility across minor versions). If shiboken does use some private Qt stuff, then that could be a problem.

This is a crucial point and definitely requires more investigation.

leycec: > If shiboken only uses/wraps the public portion of the Qt API...

I'm fairly certain that shiboken necessarily inspects both the public and private portions of the Qt API. For example, [this unresolved shiboken issue](https://bugreports.qt.io/browse/PYSIDE-434) suggests that:

> "This class [QScreen] was not implemented for a good reason: The shiboken parser is not able to understand the construction. The problem is the **initialisation of private members, which shiboken interprets wrong.**

I might just be misinterpreting that statement, but shiboken *does* seem to be breaking privacy encapsulation... probably because it's got to.

> Qt maintains backward API & ABI compatibility across minor versions

I don't quite know how to interpret this, but the [Qt ABI Tracker](https://abi-laboratory.pro/tracker/timeline/qt) *does* show a variety of symbol changes across even minor versions. Moreover, minor versions appear to occasionally introduce entirely new components. Qt 5.5.1, for example, added the new [`Qt5WaylandClient` component](https://abi-laboratory.pro/tracker/objects_report/qt/5.5.0/5.5.1/report.html).

Even if minor versions *did* perfectly preserve backward compatibility, creating Python bindings for new components added between minor versions would still require recompiling shiboken on Qt version bumps.

I think, anyway? Admittedly, I'm in over my head here. And I'm not a tall guy to begin with.

dev-python/pyside/pyside-9999.ebuild
leycec: > What I meant to say here is simply that the internal inter-dependencies among pyside modules might not be the same as those among PyQt5 modules.

**Yup!** I've already noted a few discrepancies between the two projects. *Grrrrr.* :shark:

leycec:
**Bam!** It is tragic, but we were a bit too slow on the trigger here. `shiboken2` now formally requires `clang`, presumably necessitating massive ebuild changes on my end.

This is both good (*for `clang` is good*) and bad (*for I am lazy*). Let's see where I get with this, tonight.

leycec:
**Ugh.** `shiboken2` requires Qt >= 5.7 as well now. My worst fears are blossoming before my eyes.

leycec:
Unmasking and upgrading to Qt 5.7.1 is going to take the better part of my life. In the meanwhile, I've patched this pull request with the changeset documented above.

Thanks again for the rigorous review! We'll get there... *someday.*

leycec:
**O.K.** This should be *more-or-less* ready to go now. The `shiboken2` ebuild now requires both CLang and Qt >= 5.7.1 and is successfully installable. Sadly, upstream compilation errors currently prevent the `pyside2` ebuild from being successfully installed. In theory, it should be.

Assuming upstream fixes the `pyside2` breakage in a few days, this pull request *should* be safely mergable. Let's all cross fingers and appendages.

dev-python/shiboken/shiboken-9999.ebuild
Pesa: > I might just be misinterpreting that statement, but shiboken does seem to be breaking privacy encapsulation in certain edge cases.

I think that comment just means that shiboken is not able to parse all C++ constructs and chokes on some code in the private section of that class. This does not imply that shiboken/pyside rely on the private ABI.

> Curiously, the Qt ABI Tracker shows a number of ABI symbol changes across even minor Qt versions

Only **removal** of **public** symbols breaks ABI.

> More relevantly, minor Qt versions sometimes introduce entirely new functionality. Qt 5.5.1, for example, added the new Qt5WaylandClient component.

Adding stuff doesn't break the ABI. Moreover, that module was not considered stable yet at that point in time.

> creating Python bindings for new functionality added between minor Qt versions should still require rebuilding shiboken and pyside.

Correct. In practice, it has never been an issue with PyQt. So I would not introduce a subslot operator for this reason alone.

dev-python/pyside/metadata.xml
```diff
@@ -6,17 +6,26 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
-		<flag name="X">Build QtGui and QtTest modules</flag>
+		<flag name="concurrent">Build QtConcurrent module</flag>
 		<flag name="declarative">Build QtDeclarative module</flag>
 		<flag name="designer">Build QtDesigner and QtUiTools modules</flag>
+		<flag name="gui">Build QtGui and QtTest modules</flag>
```
Pesa: QtTest? is that a typo or intentional? if the latter, what's the difference with the `testlib` flag?

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -1,71 +1,91 @@
...
-# Minimal supported version of Qt.
-QT_PV="4.8.5:4"
+# Minimum Qt version required, derived from the "CMakeLists.txt" line:
+#      find_package(Qt5 ${QT_PV} REQUIRED COMPONENTS Core)
+QT_PV="5.3.0:5"
 
 RDEPEND="
 	${PYTHON_DEPS}
-	>=dev-python/shiboken-${PV}[${PYTHON_USEDEP}]
+	>=dev-python/shiboken-${PV}:${SLOT}=[${PYTHON_USEDEP}]
```
Pesa: The `=` doesn't make sense since shiboken doesn't have a subslot

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -78,23 +98,38 @@ src_prepare() {
 src_configure() {
 	append-cxxflags -std=c++11
```
Pesa: is this still needed? C++11 is required since Qt 5.7 anyway

dev-python/shiboken/shiboken-9999.ebuild
```diff
@@ -1,44 +1,59 @@
...
+#maximum range of CLang versions. Due to insufficient upstream documentation and
+#ongoing development, this range is currently unknown.
 RDEPEND="
 	${PYTHON_DEPS}
-	dev-libs/libxml2
-	dev-libs/libxslt
-	dev-qt/qtcore:5
-	dev-qt/qtxml:5
-	dev-qt/qtxmlpatterns:5
+	sys-devel/clang:*
```
Pesa: why not `:=`? it doesn't link against clang libraries?

leycec: Exactly. From what I can tell, shiboken2 only requires the clang toolchain at compilation time:

```
$ ldd /usr/lib64/python3.4/site-packages/shiboken2.so 
	linux-vdso.so.1 (0x00007ffee56fc000)
	libpython3.4m.so.1.0 => /usr/lib64/libpython3.4m.so.1.0 (0x00007f688c186000)
	libshiboken2.cpython-34m.so.2.0 => /usr/lib64/libshiboken2.cpython-34m.so.2.0 (0x00007f688bf55000)
	libstdc++.so.6 => /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.4/libstdc++.so.6 (0x00007f688bbfe000)
	libgcc_s.so.1 => /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.4/libgcc_s.so.1 (0x00007f688b9e7000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f688b650000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f688b434000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f688b230000)
	libutil.so.1 => /lib64/libutil.so.1 (0x00007f688b02c000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f688ad31000)
	/lib64/ld-linux-x86-64.so.2 (0x0000559ed7a0e000)
```

I'm probably missing something. It is a bit eerie that shiboken2 doesn't link against clang, isn't it?

dev-python/pyside/pyside-9999.ebuild
leycec: **Great catch.** Your eyes are better than mine. It's time for a new monocle.

dev-python/pyside/pyside-9999.ebuild
leycec: Good catch. Darn those subslots to heck! This shall be fixed.

dev-python/pyside/metadata.xml
leycec: **Wow.** That is indeed a typo, presumably inherited from the prior `pyside-9999.ebuild`. Your attention to detail is impressive. I who am about to submit a working pull request salute you!

leycec:
*Done.* Thanks again for the exhaustive code review. In theory, this latest commit should address most of the above concerns – including:

* The C++11 `${CFLAG}` patch has been removed (as Qt >= 5.7.1 already mandates use of C++11).
* Numerous typos in USE flag descriptions have been revised.
* PySide2 no longer requires shiboken2 via a subslot operator (as shiboken2 has no subslot).
* shiboken2 no longer requires Qt via a subslot operator (as that is probably a bad idea).

leycec:
I've also submitted two upstream issues to the PySide2 JIRA:

* [**Optional Qt Build Dependencies Incorrectly Listed as Mandatory**](https://bugreports.qt.io/browse/PYSIDE-502), intended to make our current `sed` kludge in `pyside-9999.ebuild` go away.
* [**Compilation Fails with Terrifying Clang Error on "Running generator for QtCore..."**](https://bugreports.qt.io/browse/PYSIDE-501?filter=-2), intended to make `pyside-9999.ebuild` compile again. It currently doesn't, and I can't figure out why.

Unless the [second issue](https://bugreports.qt.io/browse/PYSIDE-501?filter=-2) is resolved next week, I'll revert both ebuilds to the Qt 5.6 branches of PySide2 and shiboken. It's a bit frustrating, honestly. The current `dev` branch is too much of a moving target for us to reliably hit, at the moment.

dev-python/shiboken/shiboken-9999.ebuild
Pesa: Interesting. What about the `shiboken` binary and the `/usr/lib64/libshiboken2*` libraries?

Pesa:
Upstream says:

> The branch that you should test against is 5.6. The dev branch is in flux and not meant for testing.

leycec:
**Right.** That's probably in response to the [second issue above](https://bugreports.qt.io/browse/PYSIDE-501). I'll rewind time back to the `5.6` branch and hope for blessed worky. :crossed_fingers:

leycec:
The promised land beckons. PySide2 and shiboken2 have both been reverted to the `5.6` branches of their respective git repositories. Unsurprisingly, both now successfully install.

To simplify migration to the `clang`-enabled `dev` branches, several ignorable `FIXME:` comments have been left rotting in both ebuilds. If this is badness, I'll rip them out. Else, this pull request should now be feature complete.

Thanks for your continued vigilance, Pesa. :crab:

Pesa:
Excellent job. Thanks!
One last question: do we really need to force Qt == 5.6.x? PyQt5 *usually* works fine against newer versions of Qt (it doesn't support any new Qt features, obviously, but it does build).
Also, minor nitpick, can you come up with a more informative commit msg title, instead of "fix bug 612272"?

leycec:
> One last question: do we really need to force Qt == 5.6.x? PyQt5 usually works fine against newer versions of Qt (it doesn't support any new Qt features, obviously, but it does build).

Excellent query. I meant to highlight this in my prior comment, but clearly dropped the ball. Thankfully, nothing gets by Pesa.

The short answer is: **Yup.** The `5.6` branch of both PySide2 and shiboken2 sadly requires Qt 5.6.x. In particular, attempting to build the `5.6` branch of shiboken2 against Qt >= 5.7 will fail fairly quickly with a spate of C++11 parser errors. That's right! The hand-rolled C++ parser bundled with both shiboken1 and this branch of shiboken2 can't parse C++11. <sup>_Kill me._</sup>

At least it reliably parses most pre-C++11 syntax. Let us pray the `clang`-enabled `dev` branch of PySide2 stabilizes soon.

> Also, minor nitpick, can you come up with a more informative commit msg title, instead of "fix bug 612272"?

You're totally right. Something this epic deserves a less pathetic title. Let's invent one.

leycec:
_Bam!_ How's that?

Pesa:
> The short answer is: "Yup." The 5.6 branch of both PySide2 and shiboken2 sadly requires Qt 5.6.x. In particular, attempting to build the 5.6 branch of shiboken2 against Qt >= 5.7 fails on the first C++11 syntax it stumbles over. That's right! The hand-rolled C++ parser bundled with both shiboken1 and this branch of shiboken2 can't parse C++11 syntax... at all. Kill me.

oh yeah you did mention this problem earlier, my bad.

Pesa:
btw we probably need a live ebuild for pyside-tools to go along with these...

Pesa:
Merged, thanks a lot!

leycec:
> btw we probably need a live ebuild for pyside-tools to go along with these...

You're right... *of course.* Why are you always right? My free time hates you.

> Merged, thanks a lot!

:fireworks: **+** :champagne: **+** :dancer:  **=** :relaxed:

Pesa:
Also, there's something very wrong with the `.cmake` files installed by shiboken.
Root cause is `PYTHON_EXTENSION_SUFFIX` being empty for python2.7 builds, thus the substitutions in `data/CMakeLists.txt` go awry.

leycec:
Perceptive catch. As you say, that's absolutely not right. I suspected there might be subtle badness given suspicious `cmake` output on installing `pyside:2` – but ultimately ignored everything, due to `PySide2` superficially appearing to work. <sup>_Perhaps that was the wrong decision._</sup>

I'll inspect that and probably submit an upstream issue. Is this significant enough to warrant a `sed` patch on our end in the meanwhile? I assume **"Gods, yes!"**, but hope for "No, just keep being lazy."

I also have a `pyside-tools:2` ebuild in the works, which I'll submit shortly.

Thanks for the `virtx` and `LICENSE` catches, too. The latter is vital. I... can't believe I missed that. :crying_cat_face:

Pesa:
> I'll inspect that and probably submit an upstream issue. Is this significant enough to warrant a sed patch on our end in the meanwhile?

No, this should go upstream. Keep being lazy ;)

leycec:
[**Issue reported.**](https://bugreports.qt.io/browse/PYSIDE-508) Thanks to your detailed analysis, I laboriously tracked the underlying issue down to the `get_python2_release_suffix()` macro in shiboken2's top-level `CMakeLists.txt`. Unsurprisingly, the macro is broken.

Up next: `dev-python/pyside-tools`. The excitement is palpable. :stadium:

PR #143, opened 2017-04-16
Update qt5-build.eclass

perfect7gentleman:
that solves https://bugs.gentoo.org/show_bug.cgi?id=599636


Pesa:
NAK

This is clearly not the correct fix.

perfect7gentleman:
Why not?

Pesa:
If we don't pass `-no-dbus` the default is autodetection. And if dbus is autodetected while building qtcore (the normal case, since dbus is installed in almost every system), this will be recorded in files generated by the build process, which will confuse other packages that have an optional qtdbus dependency.

PR #144, opened 2017-04-17
dev-qt: Add live ebuilds for the 5.9 branch.

Chiitoo:
Tested only building the packages required for QupZilla, using only one combination of USE-flags.

Copied/renamed the most current '5.9999' ebuilds, with some minor tweaks.

For 'qtcore':

- Added 'src/tools/qfloat16-tables' to 'QT5_TARGET_SUBDIRS'.
- Changed dependency from '>=dev-libs/libpcre-8.38[pcre16,unicode]' to 'dev-libs/libpcre2[pcre16,unicode]'.

For 'qtwebengine':

- Removed patches.
- Changed 'src/core/core_gyp_generator.pro' to 'src/core/core_generator.pro'.
- Removed the removal of 'src/3rdparty/chromium/third_party/icu/source'.  Have not yet really tested if the system-icu was indeed used (config did say it was).  I saw one particular commit that may or may not have something to do with this: http://code.qt.io/cgit/qt/qtwebengine.git/commit/?h=5.9&id=16d8518e58fcaf6495604ae2d9188cc59b3e661b
- Reverted http://code.qt.io/cgit/qt/qtwebengine.git/commit/?h=5.9&id=f6c1a34972eb2d9132346ab4eb442855d3ef555e since for a reason or another, the check fails to detect 'icuuc' even though it is there.

With regards to 'qtwebengine', I left only the removal of patches, and renaming of 'src/core/core_gyp_generator.pro' to 'src/core/core_generator.pro' in for this push, as it will probably be better to fix the other things in separate commits, and more proper-like.

Colour me surprised if I didn't forget anything.

### DIFF
--- a/dev-qt/assistant/assistant-5.9.9999.ebuild
+++ b/dev-qt/assistant/assistant-5.9.9999.ebuild
@@ -1,0 +1,37 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+QT5_MODULE="qttools"
+inherit qt5-build
+
+DESCRIPTION="Tool for viewing on-line documentation in Qt help file format"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~arm ~hppa ~ppc64 ~x86"
+fi
+
+IUSE="webkit"
+
+DEPEND="
+	~dev-qt/qtcore-${PV}
+	~dev-qt/qtgui-${PV}
+	~dev-qt/qthelp-${PV}
+	~dev-qt/qtnetwork-${PV}
+	~dev-qt/qtprintsupport-${PV}
+	~dev-qt/qtsql-${PV}[sqlite]
+	~dev-qt/qtwidgets-${PV}
+	webkit? ( ~dev-qt/qtwebkit-${PV} )
+"
+RDEPEND="${DEPEND}"
+
+QT5_TARGET_SUBDIRS=(
+	src/assistant/assistant
+)
+
+src_prepare() {
+	qt_use_disable_mod webkit webkitwidgets \
+		src/assistant/assistant/assistant.pro
+
+	qt5-build_src_prepare
+}
--- a/dev-qt/designer/designer-5.9.9999.ebuild
+++ b/dev-qt/designer/designer-5.9.9999.ebuild
@@ -1,0 +1,40 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+QT5_MODULE="qttools"
+inherit qt5-build
+
+DESCRIPTION="WYSIWYG tool for designing and building Qt-based GUIs"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~x86"
+fi
+
+IUSE="declarative webkit"
+
+DEPEND="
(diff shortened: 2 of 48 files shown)


Chiitoo:
Thanks.  Yeah, I had a feeling you do these like that, but wasn't sure about the details.

One question though: do the headers matter, since repoman likes to fix things like the copyright year?

Pesa:
> One question though: do the headers matter, since repoman likes to fix things like the copyright year?

Not sure what you mean...? If the copyright year is giving you headaches (e.g. noise in diffs), just send a PR to update all of them to 2017 in all live ebuilds.

Pesa:
It seems you're using different pull requests to split the changes (totally fine), so I'm going to close this one.

Chiitoo:
Repoman by default seems to update the copyright years now, so when doing just a copy of one that has a 2016 still, it's changing it to 2017 automagically.

And yeah, I was thinking of doing at least some of the changes already separately, and then re-do this one with just the copied ebuilds.

Pesa:
well the copyright year difference is not a problem, feel free to update all of them at once beforehand if they're bothering you ;)

ok, I'll reopen then.

Chiitoo:
Well, they're not bothering me, but I wanted to make sure since you mentioned that the ebuilds should be exact copies when adding them, but if I use repoman it updates the headers automagically on everything I commit (I'd guess that can be disabled).

Just using plain git commands would of course be an option, too, so the headers aren't touched.

So, by “at once beforehand”, do you mean as a separate commit as well, before the commit that adds the new ebuilds?  For new files I imagine it could make sense to have 2017... and for the rest I think in general I'd only change it when making other changes to the file.

Pesa:
Yeah don't worry, just ignore the copyright line. That one can be different. So to amend my previous request... "exact copies" *except* for the copyright year.

Chiitoo:
Gotcha!

As a sidey-question, which ebuilds do you use as a base for, say, 5.9.0_beta, and would those be added in a similar fashion: copy, then change what need be?

Pesa:
`5.9.0_betaX` should be based on `5.9.9999`. If a fix is needed *only* in 5.9.0_betaX (e.g. because the bug has been fixed upstream on the 5.9 branch *after* the beta), then the change should obviously be applied to `5.9.0_betaX` only.

Chiitoo:
OK this should be clean now, so to speak.

Would it make the most sense to also do the 5.9.0_beta ebuilds now, before implementing fixes on these (then do both as needed)?  Or rather fix the things I already have fixes for before going for it?

Pesa:
What happened to the `qtwebengine` changes you described earlier?

> Would it make the most sense to also do the 5.9.0_beta ebuilds now, before implementing fixes on these (then do both as needed)? Or rather fix the things I already have fixes for before going for it?

Either way is fine.

Chiitoo:
I figured the changes on 'qtwebengine' could wait for a bit longer, as I'm not sure about a lot of it right now.  For example, I don't know if it's a good thing to remove the ICU-workaround, though at least the build did not fail, which seems is why the workaround was originally added.  Should probably make sure the 'system-icu' is really used, somehow.

I'm also not sure if it's normal that at least some of the tests seem to be running twice.  Possibly related to the switch to 'GN' internally?  Our old ways of setting system-ffmpeg still seems to work, but the first configure summary will say no system-ffmpeg, while the second one says system and bundled unless I remove the line [39 here](http://code.qt.io/cgit/qt/qtwebengine.git/tree/tools/qmake/mkspecs/features/configure.prf?h=5.9&id=f6c1a34972eb2d9132346ab4eb442855d3ef555e#n39).

The libxml issue I've reported [upstream](https://bugreports.qt.io/browse/QTBUG-60211), so we can probably more or less forget about that.

Pesa:
Alright then.

PR #145, opened 2017-04-17
dev-qt/qtcore: Adjust 'libpcre' dep and add a subdir.

Chiitoo:
Configuring will now fail without 'dev-libs/libpcre2[pcre16]' and the new 'src/tools/qfloat16-tables' subdir needs to be processed before 'src/corelib'.


Chiitoo:
I just checked, a little late, that the minimum version for 'libpcre' appears to be 10.20 which I did not specify as I was unaware of it.

The lowest we have in Portage seems to be 10.22.  Should it still be specified?

Pesa:
Merged, thanks!

PR #146, opened 2017-04-20
Qt >= 5.8: Build fixes

BinaryKhaos:
The commit msgs should be rather self-explanatory, I hope.

Generally it fixes the compilation of qtdbus and qtgui (+dbus) for Qt >= 5.8 as well as other packages that depend on qtdbus. I used 5.8 git head for my tests and recompiled kde-frameworks/* ... everything works just fine on my machine and I could not find any negative side-effects.


Pesa:
Merged.

PR #147, opened 2017-04-24
dev-qt/qtwebengine: Fix configure for the live 5.9 ebuilds.

Chiitoo:
This renames 'src/core/core_gyp_generator.pro' to 'src/core/core_generator.pro', and removes the unneeded/not applying patches from being used.

### DIFF
--- a/dev-qt/qtwebengine/qtwebengine-5.9.9999.ebuild
+++ b/dev-qt/qtwebengine/qtwebengine-5.9.9999.ebuild
@@ -86,7 +86,8 @@ src_prepare() {
 
 	qt_use_disable_mod geolocation positioning \
 		src/core/core_common.pri \
-		src/core/core_gyp_generator.pro
+		src/core/core_chromium.pri \
+		tools/qmake/mkspecs/features/configure.prf
 
 	qt_use_disable_mod widgets widgets src/src.pro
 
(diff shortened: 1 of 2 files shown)


Chiitoo:
Actually... removed the 'src/core/core_gyp_generator.pro' completely, since the content that 'qt_use_disable_mod' was working on doesn't exist doesn't exist in the new file at all.

Pesa:
And are you sure it wasn't moved elsewhere?

Chiitoo:
Can't really say for sure at this time.  I'm definitely not able to see it if it was.

There is the 'tools/qmake/mkspecs/features/configure.prf', but it looks like that has been around before already.  I do wonder if we could make use of it though, since so far it might be the only place where hacking things off did anything to the configure results...

Pesa:
It's in `src/core/core_chromium.pri`, that needs to be added. And add `tools/qmake/mkspecs/features/configure.prf` as well.

Chiitoo:
Huh.  I think I saw 'core_chromium.pri' while grepping, but didn't really investigate.  I imagine I thought I already checked it, or something else confused me out.  :]

Thanks!

Both added.

I experimented with the 'configure.prf' file a bit more, and it does seem like we could perhaps use this same mechanism with, for example, ffmpeg.  Right now the automagic seems to enable bundled-ffmpeg as well (to some extent), in addition to system-ffmpeg (if looking at the configure summary).  What it is actually doing with the bundled one, I'm not sure, but if I use the same logic to edit a line related to bundled-ffmpeg in this particular file, it will disappear from the configure summary.

Pesa:
Excellent. Merged.

PR #148, opened 2017-04-27
dev-qt/qtgui: Fix building 5.9 with 'dbus' disabled.

Chiitoo:
The changes here are almost definitely no good as they are, but they at least show one idea of how to get 'qtgui[-dbus]' to build.

Without any changes, the compilation will fail due to 'unknown module in QT: dbus' or so.  Adding '-no-dbus' helps, but then it's looking for 'qdbusplatformmenu_p.h' (or so) because 'QT_NO_DBUS' is not defined for whatever reason, and defining it in 'src/platformsupport/themes/themes.pro' helps with that.

This is enough to make the compile a success, though I can still see something that looks like automagic picking up of 'qtdbus' if it is installed, when building other targets.

Similarly to earlier commits, the file(s) 'qt_use_disable_config' was working on has/have been changed, but I'm not sure it's of much (if any) help.

Some upstream changes that I believe are related to that in particular:

- http://code.qt.io/cgit/qt/qtbase.git/diff/?h=5.9&id=60985aa42b37217fb4c01ed85bbf6f14410c3491
- http://code.qt.io/cgit/qt/qtbase.git/diff/?h=5.9&id=13e7c603d51f22763d47e98a4e37468e46deac98

### DIFF
--- a/dev-qt/qtgui/qtgui-5.9.9999.ebuild
+++ b/dev-qt/qtgui/qtgui-5.9.9999.ebuild
@@ -123,10 +123,14 @@ src_prepare() {
 	# egl_x11 is activated when both egl and xcb are enabled
 	use egl && QT5_GENTOO_CONFIG+=(xcb:egl_x11) || QT5_GENTOO_CONFIG+=(egl:egl_x11)
 
-	use dbus || sed -i -e 's/contains(QT_CONFIG, dbus)/false/' \
-		src/platformsupport/platformsupport.pro || die
+	use dbus || sed -i -e '1 i\DEFINES += QT_NO_DBUS' \
+		src/platformsupport/themes/themes.pro || die
 
 	qt_use_disable_config tuio udpsocket src/plugins/generic/generic.pro
+
+	qt_use_disable_config dbus dbus \
+		src/src.pro \
+		src/platformsupport/themes/genericunix/genericunix.pri
 
 	qt_use_disable_mod ibus dbus \
 		src/plugins/platforminputcontexts/platforminputcontexts.pro
@@ -139,7 +143,7 @@ src_prepare() {
 
 src_configure() {
 	local myconf=(
-		$(usex dbus -dbus-linked '')
+		$(usex dbus -dbus-linked '-no-dbus')
 		$(qt_use egl)
 		$(qt_use eglfs)
 		$(usex eglfs '-gbm -kms' '')
--- a/dev-qt/qtgui/qtgui-5.9999.ebuild
+++ b/dev-qt/qtgui/qtgui-5.9999.ebuild
@@ -123,10 +123,14 @@ src_prepare() {
 	# egl_x11 is activated when both egl and xcb are enabled
 	use egl && QT5_GENTOO_CONFIG+=(xcb:egl_x11) || QT5_GENTOO_CONFIG+=(egl:egl_x11)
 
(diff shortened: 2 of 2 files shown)


dev-qt/qtgui/qtgui-5.9.9999.ebuild
```diff
@@ -123,11 +123,15 @@ src_prepare() {
...
 
-	use dbus || sed -i -e 's/contains(QT_CONFIG, dbus)/false/' \
-		src/platformsupport/platformsupport.pro || die
+	use dbus || sed -i -e '1 i\DEFINES += QT_NO_DBUS' \
+		src/platformsupport/themes/themes.pro || die
 
 	qt_use_disable_config tuio udpsocket src/plugins/generic/generic.pro
 
+	qt_use_disable_config dbus dbus \
+		src/src.pro \
```
Pesa: `src/src.pro` should not matter to qtgui, ever.

dev-qt/qtgui/qtgui-5.9.9999.ebuild
```diff
@@ -139,7 +143,7 @@ src_prepare() {
 
 src_configure() {
 	local myconf=(
-		$(usex dbus -dbus-linked '')
+		$(usex dbus -dbus-linked '-no-dbus')
```
Pesa: unwanted side effect of 53e51ab097bd73ec05d56c389bc68da17eaf22d5?

Chiitoo:
> just to be clear, 5.8.x builds fine?

Seems like 5.8.0 is already affected.

> src/src.pro should not matter to qtgui, ever.

Ah, good to know!

> unwanted side effect of 53e51ab?

Most likely since we stopped passing '-no-dbus' globally, the automagic is doing its thing.

Somewhat puzzled why 'QT_NO_DBUS' is not defined, even though configure does pick up the switch.

Chiitoo:
> Somewhat puzzled why 'QT_NO_DBUS' is not defined, even though configure does pick up the switch.

I should probably have figured this one out a lot sooner, considering kensington's comment in IRC a couple of months ago, as well as the commit c994ea4ec0e9ff27353e85ed66ed4733f6315c5d: it's of course defined by building 'qtcore' with '-no-dbus' (or '-no-feature-dbus').

Should we add a USE="dbus" to 'qtcore', and depend upon it where need be?

Pesa:
> Should we add a USE="dbus" to 'qtcore', and depend upon it where need be?

No. That's exactly what we've been trying to avoid all this time, since the modularization of qt ebuilds. If we start doing that, we might as well just merge all qtbase packages back together.

Pesa:
We need to take a step back and fix this mess properly. Qt 5.8 introduced an almost complete rewrite of the configuration system. We've been patching what we could here and there, mostly with hacks, but it's clear that this approach does not work. I don't want the ebuilds (and `qt5-build.eclass`) to grow into an unmaintainable pile of hacks. I've been there and it made adding multilib support to Qt4 much harder and time-consuming than it should have been.

Therefore, I'm going to reject this PR and start working on properly getting 5.9.0 into shape next week (no point in wasting time on 5.8.0 anymore, upstream decided there won't be any more releases from the 5.8 branch, and 5.9.0 should be out by the end of the month). Help is of course welcome, hacks are not.

Chiitoo:
Indeed, I knew there must be a proper way to fix this all, but it's beyond what I can think of.

Thank you!

PR #150, opened 2017-05-02
dev-qt/qtnetworkauth: new package (tech preview)

kenya888:
QtNetworkAuth comes from Qt 5.9 (TP), developed on 5.8 branch also.


kenya888:
travis-ci log:
>   media-gfx/phototonic/phototonic-9999.ebuild: Ebuild uses http:// but should use https:// on line: 14
mmm... what should I do?

Chiitoo:
I believe the 'media-gfx/phototonic/phototonic' issue is unrelated to your changes.

The actual issue seems to be the 'dependency.bad' hits, which in turn are due to the 5.8.0 packages being masked in 'profiles/package.mask' (I think).

Pesa:
> The actual issue seems to be the 'dependency.bad' hits, which in turn are due to the 5.8.0 packages being masked in 'profiles/package.mask' (I think).

Correct. But please forget about 5.8.x, we will be dropping those ebuilds soon and jump directly to 5.9.0

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
```diff
@@ -0,0 +1,23 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+inherit qt5-build
+
+DESCRIPTION="Network authorization component library for the Qt5 framework"
+LICENSE="GPL-3"
```
Pesa: are you sure about this?

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
```diff
@@ -0,0 +1,23 @@
...
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~x86"
+fi
+
+IUSE=""
+
+DEPEND="
+	~dev-qt/qtnetwork-${PV}
```
Pesa: missing `qtcore` dep

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
```diff
@@ -0,0 +1,23 @@
...
+"
+RDEPEND="${DEPEND}"
+
+src_prepare() {
```
Pesa: don't define `src_prepare()` if you're only calling the eclass function.

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
kenya888: Sorry, could you let me know more about your question?
It's no problem if you concern the license. qtnetworkauth is published under GPL-3 [1]

[1] http://code.qt.io/cgit/qt/qtnetworkauth.git/tree/?h=5.9

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
kenya888: Oh, OK.

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
kenya888: I see.

kenya888:
@Chiitoo Thank you for your support.
@Pesa Thank you for your review. I'd like to update the change.

kenya888:
@Pesa Is it better for us to wait pushing this change without 5.8.x files after 5.9 is officially released?

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
Pesa: yeah I was asking about the license. You're right though, qtnetworkauth seems to be GPL3-only, unlike most other qt modules.

Pesa:
> @Pesa Is it better for us to wait pushing this change without 5.8.x files after 5.9 is officially released?

No, I'm going to drop 5.8.x as soon as next week I think. People can use the 5.9.0 betas if they want.

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
kenya888: Just for a reference, qtcharts has been also published under GPL-3 only. [2]
I refer the ebuild file of qtcharts for this change.
Could you let me know if I should care more about the license?

[2] http://code.qt.io/cgit/qt/qtcharts.git/tree/?h=5.9

kenya888:
> No, I'm going to drop 5.8.x as soon as next week I think. People can use the 5.9.0 betas if they want.

OK, I continue working. Thanks.

dev-qt/qtnetworkauth/qtnetworkauth-5.9999.ebuild
kenya888: I refer the ebuild file of qtcharts, that is published under GPL-3 [2]
Could you let me know if I should care more about the license?

[2] http://code.qt.io/cgit/qt/qtcharts.git/tree/

Pesa:
Squashed, removed 5.8.x ebuilds, and merged. Thanks a lot!

kenya888:
Thanks!!

PR #151, opened 2017-05-05
PySide2-Tools: Arise from Your Slumber!

leycec:
This pull request provides the long-promised `pyside-tools:2` ebuild and minor changes to the `pyside:2` ebuild to support both `pyside-tools:2` and sanity.

## PySide2: Supporting Multiple Python 3.x Versions for Great Justice

First, thanks for simplifying the `shiboken:2` ebuild. Less is more.

This pull request simplifies the `pyside:2` ebuild similarly *and* fixes support for building against multiple Python 3.x versions. Currently, building PySide2 for multiple Python 3.x versions *only* builds against the version of Shiboken2 specific to the newest Python 3.x version.  Older Python 3.x versions of Shiboken2 are silently ignored. Why? Because life hates leycec.  <sup>_That's me._</sup>

To fix this, the `pyside:2` ebuild now explicitly sets the `PYTHON_EXTENSION_SUFFIX` CMake option to the extension suffix unique to the Python 3.x version being built against (e.g., `.cpython-34m` for Python 3.4).  Sadly, this option is also reused for shared libraries filenames generated by PySide2 and hence *cannot* be set to a convenient Gentoo-specific string (e.g., `${EPYTHON}`). Instead, this global is only safely settable by capturing and munging the output of the `"${PYTHON_CONFIG}" --extension-suffix` command for the current Python interpreter. <sup>_It sucks._</sup>

This has the unfortunate consequence of requiring CPython 3.x, but the fortunate advantage of actually working. CPython is required, because `${PYTHON_CONFIG}` is available only for CPython. CPython 3.x is required, because the `--extension-suffix` option is available only for CPython 3.x.

To enforce this requirement, `python2_7` has been temporarily stripped from the `${PYTHON_COMPAT}` global for PySide2. Once upstream resolves [issue 508](https://bugreports.qt.io/browse/PYSIDE-508), we *should* be able to devise some means of forcing the Python 2.7 version of PySide2 to build against the Python 2.7 version of Shiboken2.

We are brilliant and capable. We can do this.

## PySide2-Tools: How Bad Could It Get?

Not too bad, actually. Unlike PySide1-Tools, PySide2-Tools is sane and supports out-of-source CMake builds. This is good. Maybe life no longer hates leycec.

Like PySide2:

* CPython 3.x is required, thanks to the same [upstream issue](https://bugreports.qt.io/browse/PYSIDE-508).
* Similar ebuild machinery is needed to force PySide2-Tools to build against the versions of PySide2 and Shiboken2 specific to the current CPython 3.x version.

But it actually works.

I'll take working.

### DIFF
--- a/dev-python/pyside-tools/pyside-tools-9999.ebuild
+++ b/dev-python/pyside-tools/pyside-tools-9999.ebuild
@@ -1,0 +1,140 @@
+# Copyright 1999-2017 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+# Perform an in- rather than out-of-source CMake build, permitting
+# src_prepare() to safely patch the codebase for each enabled Python version.
+CMAKE_IN_SOURCE_BUILD="1"
+
+#FIXME: Restore Python 2.7 support after this upstream issue is resolved:
+#    https://bugreports.qt.io/browse/PYSIDE-508
+#Even after this issue is resolved, further ebuild changes will probably be
+#needed to fully support Python 2.7. Unlike PySide2 and Shiboken2, the root
+#"CMakeLists.txt" file in pyside2-tools provides no ${PYTHON_EXECUTABLE} option.
+#Instead, the ${PYTHON_EXTENSION_SUFFIX} and ${PYTHON_BASENAME} options must be
+#passed with values specific to Python 2.7. Additionally, note that the
+#"python2.7-config" command provides no "--extension-suffix" option.
+PYTHON_COMPAT=( python3_{4,5,6} )
+# PYTHON_COMPAT=( python2_7 python3_{4,5,6} )
+
+inherit cmake-utils python-r1 virtualx git-r3
+
+DESCRIPTION="PySide development tools (lupdate, rcc, uic)"
+HOMEPAGE="https://wiki.qt.io/PySide2"
+EGIT_REPO_URI=(
+	"git://code.qt.io/pyside/${PN}.git"
+	"https://code.qt.io/git/pyside/${PN}.git"
+)
+#FIXME: Switch to the clang-enabled "dev" branch once stable.
+EGIT_BRANCH="5.6"
+
+# See the top-level "LICENSE-rcc" and "LICENSE-uic" files for licensing.
+LICENSE="BSD GPL-2"
+SLOT="2"
+KEYWORDS=""
+IUSE="test"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+# Minimum version of Qt required.
+QT_PV="5.6*:5"
+
+# The "pyside2uic" package imports both the "PySide2.QtGui" and
+# "PySide2.QtWidgets" C extensions and hence requires "gui" and "widgets".
+RDEPEND="
+	${PYTHON_DEPS}
+	>=dev-python/pyside-${PV}:${SLOT}[gui,widgets,${PYTHON_USEDEP}]
+	>=dev-python/shiboken-${PV}:${SLOT}[${PYTHON_USEDEP}]
+	=dev-qt/qtcore-${QT_PV}
+"
+DEPEND="${RDEPEND}
+	test? ( virtual/pkgconfig )
+"
+
+src_prepare() {
+	# Prepare the codebase before copying.
+	cmake-utils_src_prepare
+
+	# Copy the codebase into work directories for each enabled Python version.
+	python_copy_sources
+
+	preparation() {
+		# Change to the work directory for the current enabled Python version.
+		pushd "${BUILD_DIR}" >/dev/null || die
+
+		if python_is_python3; then
+			# Remove Python 2-specific paths.
+			rm -rf pyside2uic/port_v2 || die
+
+			# Generate proper Python 3 test interfaces with the "-py3" option.
+			sed -i -e 's:${PYSIDERCC_EXECUTABLE}:"${PYSIDERCC_EXECUTABLE} -py3":' \
+				tests/rcc/CMakeLists.txt || die
+		else
+			# Remove Python 3-specific paths.
+			rm -rf pyside2uic/port_v3 || die
+		fi
+
+		# Force testing against the current Python version.
+		sed -i -e "/pkg-config/ s:shiboken2:&-${EPYTHON}:" \
+			tests/rcc/run_test.sh || die
+
+		popd >/dev/null || die
+	}
+	python_foreach_impl preparation
+}
+
+src_configure() {
+	configuration() {
+		local mycmakeargs=(
+			-DBUILD_TESTS=$(usex test)
+		)
+
+		if python_is_python3; then
+			# Extension tag unique to the current Python 3.x version (e.g.,
+			# ".cpython-34m" for CPython 3.4).
+			local EXTENSION_TAG="$("$(python_get_PYTHON_CONFIG)" --extension-suffix)"
+			EXTENSION_TAG="${EXTENSION_TAG%.so}"
+
+			mycmakeargs+=(
+				#FIXME: Submit an upstream PySide2 issue requesting that the
+				#"PySide2Config.cmake" file use the same
+				#${PYTHON_EXTENSION_SUFFIX} variable as the
+				#"Shiboken2Config.cmake" file.
+
+				# Find the previously installed "Shiboken2Config.*.cmake" and
+				# "PySide2Config.*.cmake" files specific to this Python 3.x
+				# version.
+				-DPYTHON_EXTENSION_SUFFIX="${EXTENSION_TAG}"
+				-DPYTHON_BASENAME="${EXTENSION_TAG}"
+			)
+		else
+			die "Python 2.7 currently unsupported."
+		fi
+
+		CMAKE_USE_DIR="${BUILD_DIR}" cmake-utils_src_configure
+	}
+	python_foreach_impl configuration
+}
+
+src_compile() {
+	compilation() {
+		CMAKE_USE_DIR="${BUILD_DIR}" cmake-utils_src_compile
+	}
+	python_foreach_impl compilation
+}
+
+src_test() {
+	local -x PYTHONDONTWRITEBYTECODE
+
+	testing() {
+		CMAKE_USE_DIR="${BUILD_DIR}" virtx cmake-utils_src_test
+	}
+	python_foreach_impl testing
+}
+
+src_install() {
+	installation() {
+		CMAKE_USE_DIR="${BUILD_DIR}" cmake-utils_src_install
+	}
+	python_foreach_impl installation
+}
--- a/dev-python/pyside-tools/metadata.xml
+++ b/dev-python/pyside-tools/metadata.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+  <maintainer type="project">
+    <email>qt@gentoo.org</email>
+    <name>Gentoo Qt Project</name>
+  </maintainer>
+</pkgmetadata>


dev-python/pyside/pyside-9999.ebuild
```diff
@@ -135,8 +142,27 @@ src_configure() {
 
 	configuration() {
 		local mycmakeargs=(
+			-DPYTHON_EXECUTABLE="${PYTHON}"
```
Pesa: I have a patch locally that adds this. I was going to push it tomorrow after https://codereview.qt-project.org/#/c/193590/ is merged upstream. Without that upstream change, passing *only* `-DPYTHON_EXECUTABLE` to cmake breaks the configure phase for me.

dev-python/pyside/pyside-9999.ebuild
```diff
@@ -135,8 +142,27 @@ src_configure() {
...
+
+			# Extension tag unique to the current Python 3.x version (e.g.,
+			# ".cpython-34m" for CPython 3.4).
+			local EXTENSION_TAG="$("${PYTHON_CONFIG}" --extension-suffix)"
+			EXTENSION_TAG="${EXTENSION_TAG%.so}"
+
+			mycmakeargs+=(
+				# Find the previously installed "Shiboken2Config.*.cmake" file
+				# specific to this Python 3.x version.
+				-DPYTHON_EXTENSION_SUFFIX="${EXTENSION_TAG}"
```
Pesa: I don't know why you need all this. `PYTHON_EXTENSION_SUFFIX` seems to be calculated correctly on my system (at least when the aforementioned upstream change is applied).

leycec: *Crikey.* Did I overengineer this? At least on my system, `PYTHON_EXTENSION_SUFFIX` appears to be incorrectly set when multiple Python 3.x interpreters are available due to PySide2's `CMakeLists.txt` finding the `PythonLibs` and `PythonInterp` packages in the wrong order. Oddly, Shiboken2's `CMakeLists.txt` finds these packages in the right order. Somebody made a smello typo.

But, yeah... I overengineered this. The *right* solution is to submit an upstream issue begging for a PySide2 fix. `</sigh>`

I'll revert this. Let's try again.

leycec: **[Issue up](https://bugreports.qt.io/browse/PYSIDE-511).**

dev-python/pyside/pyside-9999.ebuild
leycec: **Wild.** On my system, passing both `-DPYTHON_EXECUTABLE` and `-DPYTHON_EXTENSION_SUFFIX` suffices to find the current Python 3.x interpreter.

Admittedly, my system's probably on drugs. <sup>_Every time._</sup>

Not passing `-DPYTHON_EXECUTABLE` seems like the wrong thing to do. Or does CMake have some other means of determining the current Python 3.x interpreter? Perhaps I have no idea what I'm doing.

leycec: Ah, yes. I see. This is [issue 511](https://bugreports.qt.io/browse/PYSIDE-511) again, isn't it? The fix is trivial, so I hope for a speedy resolution.

We could temporarily resolve this on our end by also passing `-DPYTHON_LIBRARIES` and `-DPYTHON_INCLUDE_DIRS`. But, really. They just need to fix this.

leycec:
`pyside:2` has been reverted back to simplicity. Even though Python 2.7 support in PySide2 is totally broken, disabling it is silly and thus bad. Once upstream resolves [issue 511](https://bugreports.qt.io/browse/PYSIDE-511), this *should* be good to go.

Until then, we wait patiently in the darkness. :candle:

leycec:
Sweet! *You* submitted that [upstream Qt patch](https://codereview.qt-project.org/#/c/193590). What are the chances that we would both independently diagnose and solve the same problem in the same exact way? It's... kinda creepy, actually.

I've removed all `pyside:2` changes from this pull request. Since we're clones, you clearly know best. :man_scientist:

leycec:
@Pesa Thanks for the `pyside` cleanup. It needed it. Is there anything you'd like me to patch up here?

dev-python/pyside-tools/pyside-tools-9999.ebuild
```diff
@@ -0,0 +1,118 @@
...
+HOMEPAGE="https://wiki.qt.io/PySide2"
+EGIT_REPO_URI=(
+	"git://code.qt.io/pyside/${PN}.git"
+	"https://code.qt.io/git/pyside/${PN}.git"
+)
+#FIXME: Switch to the clang-enabled "dev" branch once stable.
+EGIT_BRANCH="5.6"
+
+# See the top-level "LICENSE-rcc" and "LICENSE-uic" files for licensing.
+LICENSE="BSD GPL-2"
```
Pesa: where is BSD used? I see it mentioned in LICENSE-uic but not in the actual source files...

dev-python/pyside-tools/pyside-tools-9999.ebuild
```diff
@@ -0,0 +1,118 @@
...
+KEYWORDS=""
+IUSE="test"
+REQUIRED_USE="${PYTHON_REQUIRED_USE}"
+
+# Minimum version of Qt required.
+QT_PV="5.6*:5"
+
+RDEPEND="
+	${PYTHON_DEPS}
+	dev-python/pyside:2[gui,${PYTHON_USEDEP}]
```
Pesa: where is `[gui]` required?

dev-python/pyside-tools/pyside-tools-9999.ebuild
```diff
@@ -0,0 +1,118 @@
...
+	=dev-qt/qtcore-${QT_PV}
+	=dev-qt/qtgui-${QT_PV}
```
Pesa: same question

dev-python/pyside-tools/pyside-tools-9999.ebuild
```diff
@@ -0,0 +1,118 @@
...
+				# "PySide2Config.*.cmake" files specific to this Python 3.x
+				# version.
+				-DPYTHON_EXTENSION_SUFFIX="${EXTENSION_TAG}"
+				-DPYTHON_BASENAME="${EXTENSION_TAG}"
+			)
+		else
+			die "Python 2.7 currently unsupported."
+		fi
+
+		# CMAKE_USE_DIR="${BUILD_DIR}" cmake-utils_src_configure
```
Pesa: what's this?

dev-python/pyside-tools/pyside-tools-9999.ebuild
```diff
@@ -0,0 +1,118 @@
...
+"
+DEPEND="${RDEPEND}
+	test? ( virtual/pkgconfig )
+"
+
+src_prepare() {
+	cmake-utils_src_prepare
+
+	preparation() {
+		if python_is_python3; then
```
Pesa: This whole conditional works in `pyside-tools-0.2.15-r1` because we duplicate the sources for each python impl (`python_copy_sources`) and we do an in-source build (`CMAKE_IN_SOURCE_BUILD="1"`). Neither condition is true in this ebuild, so I doubt it'll work. You might not have noticed the problem because you only ever built it for one python3 impl.

dev-python/pyside-tools/pyside-tools-9999.ebuild
leycec: That's... *a really good question.* I'll inspect this strange anomaly.

dev-python/pyside-tools/pyside-tools-9999.ebuild
leycec: `</sigh>`

The answer is: **I have no idea,** of course. This ebuild was lazily migrated from the official `pyside-tools` ebuild, which requires a similar flag (namely, `X`).

I'll inspect this. If I can't uncover anything substantial, I'll just remove `[gui]`.

dev-python/pyside-tools/pyside-tools-9999.ebuild
leycec: same answer <sup>_laziness_</sup>

dev-python/pyside-tools/pyside-tools-9999.ebuild
leycec: **Oh!** So _that's_ why the official `pyside-tools` ebuild goes to such excruciatingly intense lengths to perform an in-source build. Since that ebuild is mostly undocumented and since the default out-of-source build behaved as expected, I assumed an in-source build was no longer necessary.

I assumed incorrectly. You're absolutely right: I only built `pyside-tools` against a single Python 3.x version and failed to notice **that it's completely broken for multiple Python 3.x versions.** `</sigh>`

Awesome catch. I'll have to contemplate how to correct this. I'd rather not revert to an in-source build if I can help it. Maybe I can't help it, of course. But maybe I can.

The test-specific `sed` commands can probably be safely segregated to the `src_test()` phase. Maybe? If so, that only leaves the removal of:

* The `pyside2uic/port_v2` subdirectory when building against Python 3.
* The `pyside2uic/port_v3` subdirectory when building against Python 2.

These removals can probably be safely segregated to the `src_install()` phase. Maybe? If so, the `src_prepare()` phase could then be entirely removed and the need for an in-source build goes away.

Maybe? Let's try.

dev-python/pyside-tools/pyside-tools-9999.ebuild
leycec: Stupidity, I'm afraid. I was hacking. It was late. In short, I ate too much pizza and caffeine.

This will go away.

leycec:
*Annnnd...* thanks for the great code review! I'll get this patched up in a jiffy.

dev-python/pyside-tools/pyside-tools-9999.ebuild
Pesa: yeah, thanks

dev-python/pyside-tools/pyside-tools-9999.ebuild
Pesa: > The test-specific sed commands can probably be safely segregated to the src_test() phase. Maybe?

Not sure how that can be done without copying the sources though... but I'll comment when I see the code ;)

> These removals can probably be safely segregated to the src_install() phase. 

Yeah that should work, hopefully.

leycec: > Not sure how that can be done without copying the sources though... but I'll comment when I see the code ;)

O.K., I now intuit what you mean. Assuming the `test` subdirectory is never installed (...*it doesn't appear to be*), this could probably be done by:

. In the `src_prepare()` phase, preserving a copy of the original `${S}/tests/rcc/CMakeLists.txt` and `${S}/tests/rcc/run_test.sh` files to `${T}`.
. In the `src_test()` phase:
  . Copying these files from `${T}` back to their original paths in `${S}`.
  . Performing these `sed` operations.

It's clumsy. I know. But, it *should* work. The question is: is it any better than the out-of-source build performed by the original `pyside-tools` ebuild? I'd say "...probably," but I'll leave that judgement call to you.

What do you think? Is this worth attempting *or* should I just revert back to the original out-of-source approach?

Pesa: yeah I'd say that's just too confusing for most ebuild readers... stick to the standard approach of copying the sources... It's not that pyside-tools has a large source tree anyway...

leycec: Sadly, that makes sense. Reader confusion is too heavy a price to pay. We'll do this the Right Way™, as you suggest.

leycec:
**Patched for glory.** Changes include:

* Surprisingly, the `[gui]` USE flag *is* actually required. In fact, so is `[widgets]`! The `pyside2uic` package repeatedly imports both the `PySide2.QtGui` and `PySide2.QtWidgets` C extensions. This makes sense, when I thought about it: `pyside2uic` converts XML-formatted `.ui` files into Python classes leveraging those C extensions. `</shrug>`
* Redundant dependencies (e.g., `=dev-qt/qtcore:5`, `=dev-qt/qtgui:5`) have been removed. Simplicity, you win again.
* Reversion back to an in- rather than out-of-source CMake build. Unsurprisingly, you were right. The resulting code is much easier to read and maintain. :thumb:

Due to laziness and incompetence, I've left the existing `LICENSE="BSD GPL-2"` as is. It's undeniably strange that every source file starts with the GPL v2 preamble and fails to reference the BSD license. But the `LICENSE-uic` file *does* emphatically state that:

```
The software module is double-licensed under the revised BSD and GPL
licenses.
```

The BSD license then follows. I'm inclined to believe that they mean it. Maybe? While I *could* post an upstream issue requesting clarification, upstream devs are pretty pressed for time. I'd feel bad if I wasted any more of it.

Pesa:
> Surprisingly, the [gui] USE flag is actually required. In fact, so is [widgets]! The pyside2uic package repeatedly imports both the PySide2.QtGui and PySide2.QtWidgets C extensions.

Uhm.. where? The only imports I see are for the "preview" action that nobody (certainly not the build process of another package) uses. I'm not sure I want to force a gui+widgets dep on everyone just for that. And in any case it would be a runtime-only dep. But please correct me if I'm wrong, I might have missed something.

Pesa:
> Redundant dependencies (e.g., =dev-qt/qtcore:5, =dev-qt/qtgui:5) have been removed.

`qtcore` is required explicitly from `CMakeLists.txt` though, don't rely on transitive dependencies for that. `shiboken` should be added too.

leycec:
Fair enough! I often opt for minimalism, contrary to Guido's insistence that ["explicit is better than implicit."](https://www.python.org/dev/peps/pep-0020). But Guido's always right.

I'll revert this back to explicit dependencies.

leycec:
**Reverted back to explicit dependencies.** How's that?

Pesa:
Thanks!

leycec:
Thank *you*. Ping me if you'd like any additional PySide2 help. If nothing else, I'll probably drop another pull request when upstream stabilizes a `clang`-enabled PySide2 branch.

Until then, copious beers for all. :beers:

PR #153, opened 2017-05-17
Qt5: add 'examples' USE-flag

iDarkTemplar:
Made based on https://github.com/gentoo/gentoo/pull/4541 and review comments


eclass/qt5-build.eclass
```diff
@@ -211,6 +221,10 @@ qt5-build_src_prepare() {
 		# Don't add -O3 to CXXFLAGS (bug 549140)
 		sed -i -e '/CONFIG\s*+=/ s/optimize_full//' \
 			src/{corelib/corelib,gui/gui}.pro || die "sed failed (optimize_full)"
+
+		# Don't check for cmake cache for installing example sources
+		# This is needed in order to install sources of examples for packages which build only some subdirectories out of whole package
+		sed -i -e 's/:!isEmpty(_QMAKE_CACHE_)//' mkspecs/features/qt_example_installs.prf || die "sed failed (qt_example_installs.prf)"
```
Pesa: can you explain this a bit more please? (here as a comment, not in the code)

iDarkTemplar: There is a file ${S}/mkspecs/features/qt_example_installs.prf. It's installed with dev-qt/qtcore:5.

There is a condition on current lines 30-31 for qt-5.6.2:

```
probase = $$relative_path($$_PRO_FILE_PWD_, $$dirname(_QMAKE_CONF_)/examples)
!isEmpty(probase):!contains(probase, ^\\..*):!isEmpty(_QMAKE_CACHE_) {
```

When this condition is true, qmake assumes it processes examples projects, and installs project's sources in addition to usual stuff (i.e. compiled binaries and other explicitely installed files).

When qmake is processing project file, it gets current directory relative to "${S}/examples" and saves it into "probase" variable. After that it ensures that it's value is not empty and is actually a subdirectory of "${S}/examples", i.e. contains no directions to upper directories (double dots, "..").

Last part of condition is check that variable _QMAKE_CACHE_ is not empty. It's not empty when qmake ran on top-level build directory (i.e. ${S}), which currently isn't done for some of packages: some packages only build specific subdirectories from whole qt5 module. Even though only some subdirectories are built for all of qtbase-based ebuilds (i.e. for qtcore, qtgui, and so on), due to the way it is built, variable _QMAKE_CACHE_ is always non-empty for them, but for other modules it may be empty. So, for some of packages this condition isn't true even while qmake is processing project files of examples, and it doesn't install sources of examples when it should.

Here's the list of current packages which wouldn't install sources of examples (but would still install binaries and other files) without this change:

> dev-qt/assistant
> dev-qt/designer
> dev-qt/linguist
> dev-qt/qthelp
> dev-qt/qtlocation
> dev-qt/qtpositioning

And following is the list of all packages which might be affected if they ever install examples:
```
/usr/portage/dev-qt $ grep -l QT5_TARGET_SUBDIR . -Ir | while read file ; do grep -L 'QT5_MODULE="qtbase"' $file ; done | sed -e 's:/: :g' | awk '{print $2}' | sort -u | while read file ; do echo dev-qt/$file ; done
dev-qt/assistant
dev-qt/designer
dev-qt/linguist
dev-qt/linguist-tools
dev-qt/pixeltool
dev-qt/qdbus
dev-qt/qdbusviewer
dev-qt/qdoc
dev-qt/qtdiag
dev-qt/qthelp
dev-qt/qtlocation
dev-qt/qtpaths
dev-qt/qtplugininfo
dev-qt/qtpositioning
```

There are other ways possible to fix this issue, i.e. run qmake on '${S}' which might be not desired, or to make some hack to make qmake think that it ran on '${S}', and although it's currently easy to do (I think creating empty "${S}/.qmake.cache" might be enough), I think it would rely on qmake internal implementation details too much, and I didn't check other side-effects which might be introduced while doing it.

Oh, and I just noticed a typo I made. It should be 'qmake cache', not 'cmake cache' in the comment.

iDarkTemplar:
I've rebased my changes onto 5.9.1 and moved all examples from qtbase module into separate package 'qtcore-examples'. This move would allow to get rid of circular dependencies between qtbase-based packages on qt updates. Any news on merging this pull request?

Pesa:
Thanks. I haven't forgotten about this, I just haven't had time to closely look at Qt 5.9 yet.

I'm wondering why the dependency on `qtcore-examples` is needed in other qtbase packages. Aren't examples standalone?

iDarkTemplar:
Before I moved qtbase examples into a separate package, those packages based on qtbase had this USE-flag, and I decided to keep it.

I see that CI doesn't like my new commits. I've ran repoman locally and got same results. The 'dependency.bad' issue disappears with '--without-mask' flag, not sure what exactly it doesn't like there. As for 'REQUIRED_USE.syntax', the USE-flag 'examples' is added by eclass qt5-build for these packages, and I think this is a false alarm too.

Pesa:
> Before I moved qtbase examples into a separate package, those packages based on qtbase had this USE-flag, and I decided to keep it.

so *all* qtbase examples are in `qtcore-examples`, not just qtcore ones? then why did you call it qt*core*-examples? :)

> the USE-flag 'examples' is added by eclass qt5-build for these packages

Not sure if I like that, it forces declaring `QT5_EXAMPLES_SUBDIRS` before the inherit line, which is ugly. And handling of qtbase and non-qtbase packages is completely different, which is also ugly.

The `dependency.bad` is due to `qtcore-examples` being unmasked while all other 5.9.1 ebuilds are masked.

iDarkTemplar:
> so all qtbase examples are in `qtcore-examples`, not just qtcore ones? then why did you call it qtcore-examples? :)

Yeah, all qtbase examples are in that package, that was the point of creating it. This way I get rid of cyclic dependencies between qtbase submodules when building examples. There's no point in creating similar example packages for other Qt5 modules since there are no cyclic dependencies. In my opinion it'd only bloat package count and serve no other purpose. As for package name, it can be renamed to anything.

> Not sure if I like that, it forces declaring `QT5_EXAMPLES_SUBDIRS` before the inherit line, which is ugly. And handling of qtbase and non-qtbase packages is completely different, which is also ugly.

I preferred to make 1 separate variable which enabled optional build of examples when declared (QT5_EXAMPLES_SUBDIRS), but I think it can also be implemented just by declaring USE-flag "examples" and updating QT5_TARGET_SUBDIRS in pkg_setup() phase, like it's done in qtnetwork, for example.

Handling of qtbase and non-qtbase is already really different, even without examples. Unlike most other modules, qtbase is split into a lot of packages.

iDarkTemplar:
Closing since it's outdated.

PR #154, opened 2017-05-20
dev-qt/qtwebkit: Patch compiling '5.9.*' against ICU >59.

Chiitoo:
Upstream bug: https://bugreports.qt.io/browse/QTBUG-60532

I imagine this might get fixed upstream sooner, rather than later, but I figured I'd still offer.  :]

Test compiled 5.9.9999 and 5.9.0_beta3.

### DIFF
--- a/dev-qt/qtwebkit/files/qtwebkit-5.9.0-icu-59.patch
+++ b/dev-qt/qtwebkit/files/qtwebkit-5.9.0-icu-59.patch
@@ -1,0 +1,51 @@
+# Fix building with ICU >=59.
+# Upstream bug: https://bugreports.qt.io/browse/QTBUG-60532
+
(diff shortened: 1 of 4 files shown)


Chiitoo:
Gah.

I completely forgot to make a condition for 'if' ICU 59 or higher is installed...  I don't know if things will still build with ICU <59 with these changes applied.

I had it in my mind at some point, but apparently it disappeared on the way here...

Chiitoo:
Made patching conditional depending upon the installed ICU version.

Chiitoo:
Rumour has it, that the patch from https://bugs.gentoo.org/618644 which is for 5.7.1, works for at least some of '5.9.*' as well, though I haven't personally tried other than applying for 5.9.9999.

It's somewhat different, but I can't tell if either one is better than the other.

Dessa:
this landed in 5.9, 5.9.1 and dev.

Chiitoo:
Indeed, thanks!

Closing.

PR #156, opened 2017-06-06
dev-qt/qtlocation: Fix build for 5.9.0+

jktjkt:
Because upstream has removed the top-level ``3rdparty.pro`` file (upstream commit qt/qtlocation@c54ee74acdb9757989004005baf79e99be4c9417 which is present in all 5.9.x including betas), we have to adapt and pick and choose the individual components now.

Tested with 5.9.9999.


kensington:
Thanks, picked as 735c3426e9f47ed5e77e3b340e0608136289666f.

PR #160, opened 2017-08-27
Fix bug #620348 - dev-qt/qtgui-5.9.0[-dbus] fails to build - Project ERROR: Unknown module(s) in QT_PRIVATE: dbus 

kensington:
This basically replicates some stuff that was removed upstream in 5.8 that we relied on to make split qtbase work.

### DIFF
--- a/dev-qt/qtcore/qtcore-5.10.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.10.9999.ebuild
@@ -42,3 +42,25 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+
+	local flags=(
+			ALSA CUPS DBUS EGL EGLFS EGL_X11 EVDEV FONTCONFIG FREETYPE
+			HARFBUZZ IMAGEFORMAT_JPEG IMAGEFORMAT_PNG LIBPROXY MITSHM
+			OPENGL OPENSSL OPENVG PULSEAUDIO SHAPE SSL TSLIB XCURSOR
+			XFIXES XKB XRANDR XRENDER XSYNC ZLIB
+	)
+
+	for flag in ${flags[@]}; do
+		cat >> "${D%/}"/${QT5_HEADERDIR}/QtCore/qconfig.h <<- _EOF_ || die
+
+			#if defined(QT_NO_${flag}) && defined(QT_${flag})
+			# undef QT_NO_${flag}
+			#elif !defined(QT_NO_${flag}) && !defined(QT_${flag})
+			# define QT_NO_${flag}
+			#endif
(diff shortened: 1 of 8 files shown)


kensington:
@Pesa, any thoughts? AFAIK this fixes the two main 5.9 blockers. With these fixed, I think we can push 5.9.2 to the main repo once it arrived.

dev-qt/qtcore/qtcore-5.10.9999.ebuild
```diff
@@ -42,3 +42,25 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+
+	local flags=(
+			ALSA CUPS DBUS EGL EGLFS EGL_X11 EVDEV FONTCONFIG FREETYPE
```
Pesa: why so much indentation here?

dev-qt/qtwebengine/qtwebengine-5.9.1.ebuild
```diff
@@ -74,6 +74,9 @@ DEPEND="${RDEPEND}
 src_prepare() {
 	use pax_kernel && PATCHES+=( "${FILESDIR}/${PN}-5.9.0-paxmark-mksnapshot.patch" )
 
+	# bug 620444 - ensure local headers are used
+	find "${S}" -name "*.pr*" | xargs sed -i -e 's|INCLUDEPATH += |&$$QTWEBENGINE_ROOT/include |' src/webengine/webengine.pro || die
```
Pesa: what's the last `src/webengine/webengine.pro`? isn't it matched by the find?

I think you can use `*.pr[fio]` instead of `*.pr*`, and add `-type f`, but it's not a big deal.

kensington:
These are definitely hacks, but having spent a very substantial amount of time investigating these particular issues I really don't see any other option.

The qtbase hacks essentially just replicate what was removed upstream in 5.8. The forwarding header is something that may been an oversight and could get landed upstream, but I very much doubt they would accept the ifdefs since as we know splitting qtbase like we do is very much unsupported.

There is an upstream bug for the qtwebengine issue but as usual build system issues like this are at the bottom of the priority list (if even supported at all, I've seen a few times discussion about if they support building in a root where another Qt is already installed).

So, as usual thanks for your review and pointing out my glaring errors. :-) I'll fix up these commits per your feedback, land them in master, and barring any further major issues I'll push 5.9.2 to the main repo in ~arch once it's released and looks OK (unless you have some objections of course).

kensington:
Rebased as d932d9210753c0c4143330c57b8800409f866e40...95b94093e5a8aef05a777bbec859af86d91cd194.

Pesa:
Sounds good, no objections at all.

PR #161, opened 2017-09-20
Fixes for the Qt 5.9 git branch

jktjkt:
Plkease check the commit changelog to evaluate their feasibility. I'm unsure about that fix which was originally done for D-Bus, but then it appears to still work on my system...

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -675,12 +675,6 @@ qt5_base_configure() {
 	"${S}"/configure "${conf[@]}" || die "configure failed"
 
 	popd >/dev/null || die
-
-	if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
-		# a forwarding header is no longer created since 5.8, causing the system
-		# config to always be used. bug 599636
-		cp src/corelib/global/qconfig.h include/QtCore/ || die
-	fi
 }
 
 # @FUNCTION: qt5_qmake
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -193,10 +193,6 @@ qt5-build_src_prepare() {
 		sed -i -e "/^QMAKE_CONF_COMPILER=/ s:=.*:=\"$(tc-getCXX)\":" \
 			configure || die "sed failed (QMAKE_CONF_COMPILER)"
 
-		# Respect toolchain and flags in config.tests
-		find config.tests/unix -name '*.test' -type f -execdir \
-			sed -i -e 's/-nocache //' '{}' + || die
-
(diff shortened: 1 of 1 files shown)


eclass/qt5-build.eclass
```diff
@@ -676,9 +676,10 @@ qt5_base_configure() {
 
 	popd >/dev/null || die
 
-	if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
+	if [[ ${QT5_MINOR_VERSION} -ge 8 && ${QT5_MINOR_VERSION} -lt 9 ]]; then
```
Pesa: this is always false, we don't have any 5.8 ebuilds, so you may as well delete the whole if.

jktjkt: Ack, done.

eclass/qt5-build.eclass
```diff
@@ -675,12 +675,6 @@ qt5_base_configure() {
 	"${S}"/configure "${conf[@]}" || die "configure failed"
 
 	popd >/dev/null || die
-
-	if [[ ${QT5_MINOR_VERSION} -ge 8 ]]; then
```
kensington: Removing this block will break 5.9.1. I'm checking now to see what changed upstream in 5.9.9999

kensington: Live ebuilds are built out-of-source while release are not which is triggering the failure. So, we should keep this block but move it up up slightly to still operate within ``QT5_BUILD_DIR``.

eclass/qt5-build.eclass
```diff
@@ -193,10 +193,6 @@ qt5-build_src_prepare() {
 		sed -i -e "/^QMAKE_CONF_COMPILER=/ s:=.*:=\"$(tc-getCXX)\":" \
 			configure || die "sed failed (QMAKE_CONF_COMPILER)"
 
-		# Respect toolchain and flags in config.tests
-		find config.tests/unix -name '*.test' -type f -execdir \
```
kensington: The upstream commit dropping the last references to -nocache is only 5.8 and later - should we make this conditional, or is it fine to nuke completely?

eclass/qt5-build.eclass
Pesa: good catch

eclass/qt5-build.eclass
Pesa: yep, should be made conditional.

Pesa:
Commit message titles should start with `qt5-build.eclass:`

kensington:
Thanks, fiddled with and picked as e44c9159b09912d55c2029d2dce478aba3ca4888..66dcf031932854324c8fbe1be883e66a66988968.

PR #163, opened 2017-12-10
qt5-build.eclass: fix respecting toolchain

kensington:
Bug: https://bugs.gentoo.org/639494

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -213,6 +213,11 @@ qt5-build_src_prepare() {
 		# Don't add -O3 to CXXFLAGS (bug 549140)
 		sed -i -e '/CONFIG\s*+=/ s/optimize_full//' \
 			src/{corelib/corelib,gui/gui}.pro || die "sed failed (optimize_full)"
+
+		# Respect build variables in configure tests (bug #639494)
+		if [[ ${QT5_MINOR_VERSION} -ge 9 ]]; then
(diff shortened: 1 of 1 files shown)


eclass/qt5-build.eclass
```diff
@@ -689,6 +694,33 @@ qt5_base_configure() {
 
 }
 
+# @FUNCTION: qt5_qmake_args
+# @INTERNAL
+# @DESCRIPTION:
+# Helper function to get the various toolchain-related variables.
+qt5_qmake_args() {
```
kensington: Ideally we'd call this from ``qt5_qmake`` too but I can't get the quoting to work correctly.

eclass/qt5-build.eclass
```diff
@@ -213,6 +213,11 @@ qt5-build_src_prepare() {
 		# Don't add -O3 to CXXFLAGS (bug 549140)
 		sed -i -e '/CONFIG\s*+=/ s/optimize_full//' \
 			src/{corelib/corelib,gui/gui}.pro || die "sed failed (optimize_full)"
+
+		# Respect build variables in configure tests (bug #639494)
+		if [[ ${QT5_MINOR_VERSION} -ge 9 ]]; then
+			sed -i -e "s|\"\$outpath/bin/qmake\" \"\$relpathMangled\" -- \"\$@\"|& $(qt5_qmake_args) |" configure || die
```
kensington: I think we could achieve the same thing by passing those args to ``configure`` in ``qt5_base_configure`` but I can't get the quoting to work correctly.

kensington:
Landed as c6ec22f69391bcb6d8632a6888c3a8f3b2a415ad.

PR #164, opened 2017-12-11
dev-qt/qtwebengine-dicts: initial packaging

jsbronder:
Packaging of dictionaries suitable for use by QtWebEngine per
https://doc.qt.io/qt-5/qtwebengine-webenginewidgets-spellchecker-example.html.
This is simply done by by running qwebengine_convert_dict on each
myspell dictionary.

Note that:
    - qt5-build is not used as the dictionaries are not directly
    dependent upon the version of Qt being used.
    - Dictionaries are built for every installed myspell dictionary.
    This could result in a few extra dictionaries if the user is
    stripping L10N_* flags but that saves complexity in the ebuild as
    the map from language to dictionary files(s) isn't something that
    can be simply done with string parsing.  See comments in ebuild.

Package-Manager: Portage-2.3.13, Repoman-2.3.3


kensington:
LGTM

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+		de-1901)	dict="de_1901";;
+		pt-BR)		dict="pt-br";;
+		*)			dict="${lang}";;
+	esac
+	DEPEND+=" l10n_${lang}? ( app-dicts/myspell-${dict} )"
+done
+unset dict lang
+
+RDEPEND="
+	${DEPEND}
```
Pesa: you can put this all on one line

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+	# 	en: en_AU, en_CA, en_GB, en_US, en_ZA
+	# 	kk: kk_KZ, kk_noun_adj, kk_test
+	# 	pt-BR: pt_BR
+	# etc.
+	for dic in "${EPREFIX}"/usr/share/hunspell/*.dic; do
+		bdic_fn=$(basename ${dic})
+		bdic_fn=${bdic_fn%.dic}.bdic
+		bdic_fn=${bdic_fn/_/-}
+
+		"${EPREFIX}"/usr/$(get_libdir)/qt5/bin/qwebengine_convert_dict \
```
Pesa: use `qt5_get_bindir` from `qmake-utils.eclass`

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+
+src_configure() {
+	:
+}
+
+src_compile() {
+	local dicts
+	local dic
+	local bdic_fn
+	local prefix
```
Pesa: you can declare multiple vars in one line with `local`

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+DESCRIPTION="Hunspell dictionaries for QtWebEngine"
+HOMEPAGE="https://doc.qt.io/qt-5/qtwebengine-webenginewidgets-spellchecker-example.html"
+SRC_URI=""
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~x86"
+IUSE=""
+
+S="${WORKDIR}"
```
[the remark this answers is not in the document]
jsbronder: Sure, moved.

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+
+# Cribbed from app-text/hunspell
+LANGS="af bg ca cs cy da de de-1901 el en eo es et fo fr ga gl he hr hu ia id
+is it kk km ku lt lv mi mk ms nb nl nn pl pt pt-BR ro ru sk sl sq sv sw tn uk
+zu"
+
+DEPEND="
+	app-dicts/myspell-en
+	dev-qt/qtwebengine"
```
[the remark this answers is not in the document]
jsbronder: Done

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
```diff
@@ -0,0 +1,78 @@
...
+
+for lang in ${LANGS}; do
+	IUSE+=" l10n_${lang}"
+	case ${lang} in
+		de-1901)	dict="de_1901";;
+		pt-BR)		dict="pt-br";;
+		*)			dict="${lang}";;
```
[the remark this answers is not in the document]
jsbronder: Switched to spaces, the tabs looked deceivingly fine to me.

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
jsbronder: Done

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
jsbronder: Sure, but I find that makes things less readable.  A number of style guides recommend against it in various languages.  Still, I'll change them if you insist.

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
jsbronder: Done.

jsbronder:
Regarding changing PV to a date, I don't think that makes sense here.  The eclass-manpages have a date as they're generated from a snapshot of the portage tree.  This is using whatever hunspell dictionaries are installed which cannot be directly tied to a specific date (aside from whenever the user ran emerge).

Hence I think that the 1.0 is appropriate as this is the 1.0 version of building dictionaries that qtwebengine can use.  If the packaging changes in the future, then I'd change the PV at that point.

dev-qt/qtwebengine-dicts/qtwebengine-dicts-1.0.ebuild
Pesa: No, I'm fine either way.

Pesa:
> This is using whatever hunspell dictionaries are installed which cannot be directly tied to a specific date (aside from whenever the user ran emerge).

I'm aware of that. And technically it's a problem. If the hunspell dicts are updated, the webengine dicts are not, unless the user knows that they have to manually re-emerge `qtwebengine-dicts`. How about using `PV = 5.x`, so that the dicts are refreshed at least once every qt minor version upgrade?

Pesa:
Also, should there be a dependency on this package from `qtwebengine`?

jsbronder:
Yeah, I don't have a good way to fix that.  We can make the PV match the current Qt release as you suggest.  Should there also be something in the logs/einfo about it?

I don't think that qtwebengine needs a strict dependency but it could be a USE-flag controlled one that defaults to enabled?

jsbronder:
Alternatively, we could see if getting subslots on the dictionaries is a possibility.

Pesa:
> Yeah, I don't have a good way to fix that. We can make the PV match the current Qt release as you suggest.

Yeah let's do that please.

> Should there also be something in the logs/einfo about it?

Up to you, I'm fine either way.

> I don't think that qtwebengine needs a strict dependency but it could be a USE-flag controlled one that defaults to enabled?

`USE=spell`? and no need to enable it by default.

> Alternatively, we could see if getting subslots on the dictionaries is a possibility.

I thought about that too but it doesn't seem worth the effort.

PR #166, opened 2018-02-22
Fixes for dev-qt/qtgui-5.{9,10}* w/ mesa-18 and >=dev-qt/qtsql-5.9.9999 w/ mariadb-10.2

a17r:
See also https://github.com/gentoo/gentoo/pull/7253


dev-qt/qtgui/qtgui-5.10.9999.ebuild
```diff
@@ -71,6 +71,8 @@ PDEPEND="
 	ibus? ( app-i18n/ibus )
 "
 
+PATCHES=( "${FILESDIR}/${PN}-5.9.4-opengl.patch" ) # QTBUG-66420
```
Pesa: IIRC upstream closed the 5.10 branch (no more releases from it), so we should just kill 5.10.9999 instead of wasting time backporting fixes to it.

dev-qt/qtsql/qtsql-5.10.9999.ebuild
```diff
@@ -28,6 +28,8 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
+PATCHES=( "${FILESDIR}/${PN}-5.9.5-mariadb-10.2.patch" ) # pending upstream
```
Pesa: ditto

a17r:
Also queued up my other changes over from gentoo.git

Arfrever:
Maybe these backports could be added also to `qtgui-5.11.0_alpha-r1` and `qtsql-5.11.0_alpha-r1`?

Pesa:
> Maybe these backports could be added also to qtgui-5.11.0_alpha-r1 and qtsql-5.11.0_alpha-r1?

Up to @a17r. Personally I could care less about the alpha(s), I don't think we should ever bother with them and I don't think they should've been added in the first place. Until the beta(s), users should use 5.x.9999 if they want bleeding edge.

Chiitoo:
> I don't think we should ever bother with them and I don't think they should've been added in the first place.

Sorry about that!

I should have kept them local, or at least ask around if they'd be wanted in the overlay.

Pesa:
Don't worry about that. I'm totally fine with keeping them if someone (you in this case) is willing to maintain them. My point is that I'm not going to ask @a17r to do extra work to fix/improve some short-lived alpha-quality packages that will become obsolete in a few weeks.

a17r:
In kde overlay, pre-release versions are usually masked. The same should perhaps be done here as well. Now that they're here, I just added the patches to the existing ebuilds though.

@pesa, want me to merge?

Pesa:
yes please, thank you!

PR #168, opened 2018-04-11
Qt4--

a17r:


app-crypt/qca/qca-9999.ebuild
```diff
@@ -31,80 +26,56 @@ RDEPEND="
...
-	)
 	sasl? ( dev-libs/cyrus-sasl:2 )
 	ssl? (
 		!libressl? ( >=dev-libs/openssl-1.0.1:0= )
 		libressl? ( dev-libs/libressl:= )
 	)
 "
-DEPEND="${RDEPEND}
+DEPEND="${COMMON_DEPEND}
+	dev-qt/qtnetwork:5
```
Pesa: build-time dep only?

a17r: actually, it seems only to be required with `USE=test`, going to fix upstream first. Otherwise it built fine without finding Network

PR #171, opened 2018-06-12
Try to fix global repoman/CI errors

a17r:


eclass/qmake-utils.eclass
```diff
@@ -16,16 +17,16 @@
 if [[ -z ${_QMAKE_UTILS_ECLASS} ]]; then
 _QMAKE_UTILS_ECLASS=1
 
-[[ ${EAPI:-0} == [012345] ]] && inherit eutils multilib
+[[ ${EAPI:-0} == [01234] ]] && die "qmake-utils.eclass: unsupported EAPI=${EAPI:-0}"
```
Pesa: did you check all consumers? including those in gentoo.git?

eclass/qmake-utils.eclass
```diff
@@ -16,16 +17,16 @@
...
+
+[[ ${EAPI:-0} == [5] ]] && inherit eutils multilib
```
Pesa: no need for `[` `]` around 5

eclass/qt5-build.eclass
```diff
@@ -631,7 +631,7 @@ qt5_base_configure() {
 		-no-compile-examples
 
 		# disable rpath on non-prefix (bugs 380415 and 417169)
-		$(usex prefix '' -no-rpath)
+		$(has prefix ${IUSE} && $(usex prefix '' -no-rpath))
```
Pesa: I'm confused, what does it mean if `prefix` is not in IUSE? in any case, the logic seems wrong, we should always pass `-no-rpath` unless we're on prefix.

Also, use `in_iuse` instead of `has`.

eclass/virtualx.eclass
```diff
@@ -0,0 +1,260 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+# @ECLASS: virtualx.eclass
```
Pesa: why is this necessary? what issue(s) does it fix?

eclass/qmake-utils.eclass
a17r: Yes, there was only one left which I removed. https://qa-reports.gentoo.org/output/eapi-per-eclass/qmake-utils.eclass/

eclass/virtualx.eclass
a17r: For quite some time now, repoman and CI both spew out the following error for each instance of inheriting virtualx.eclass (interestingly, not for the tree):

`DEPEND: USE flag 'prefix' referenced in conditional '!prefix?' is not in IUSE`

eclass/qt5-build.eclass
a17r: ah yes, too fast there. `!use_if_iuse` seems useful.

eclass/qmake-utils.eclass
a17r: kde-sunset.git already contains copies of Qt4 ebuilds and qt4*eclass, current state of qmake-utils.eclass in tree should also be copied there before syncing overlay changes.

eclass/qt5-build.eclass
a17r: From what I understood, the fact that Portage does not throw an error if `prefix` (among several others, such as elibc_glibc, kernel_linux) is used without definition, is wrong according to PMS.

eclass/qmake-utils.eclass
a17r: (qt.git and kde.git have already banned <6, so I did not have to check)

eclass/qt5-build.eclass
```diff
@@ -631,7 +631,7 @@ qt5_base_configure() {
 		-no-compile-examples
 
 		# disable rpath on non-prefix (bugs 380415 and 417169)
-		$(usex prefix '' -no-rpath)
+		$(!in_iuse prefix || !use_if_iuse prefix && echo -no-rpath)
```
Pesa: honestly, this is unreadable :P
Doesn't `use_if_iuse` include the `in_iuse` logic?
Also, `use_if_iuse` is defined in eutils.eclass right? we don't inherit that eclass...
I would just stick to `in_iuse`+`usex`.

a17r: can't comfortably do `usex` if `-no-rpath` shall be the default for `!in_iuse` as well as `!use_if_iuse` cases. but here's a new iteration.

eclass/qt5-build.eclass
```diff
@@ -631,7 +631,7 @@ qt5_base_configure() {
 		-no-compile-examples
 
 		# disable rpath on non-prefix (bugs 380415 and 417169)
-		$(usex prefix '' -no-rpath)
+		$(if in_iuse prefix && use prefix; then echo ''; else echo -no-rpath; fi)
```
Pesa: which can be shortened to `{ in_iuse prefix && use prefix; } || echo -no-rpath`, right?

eclass/virtualx.eclass
Pesa: this has been fixed in portage/repoman

a17r:
Thanks!

PR #172, opened 2018-06-12
Drop obsolete Qt 5.10.1

a17r:


dev-qt/qtwebengine/qtwebengine-5.11.0.ebuild
```diff
@@ -37,7 +37,7 @@ RDEPEND="
 	media-libs/libpng:0=
 	>=media-libs/libvpx-1.5:=[svc]
 	media-libs/libwebp:=
-	media-libs/mesa
+	media-libs/mesa[egl]
```
Pesa: live ebuilds not affected?

a17r: those are fine already

Pesa: oh yes, I missed that. Then I'm not sure why you need to apply this to 5.11.0 as a separate commit. When bumping to 5.11.0 you're supposed to copy from the 5.11.9999 ebuilds, which already contain the fix.

PR #176, opened 2018-06-16
Qt 5.11.1 bump

a17r:
See also: https://bugreports.qt.io/browse/QTBUG-68224

Package-Manager: Portage-2.3.40, Repoman-2.3.9


Pesa:
does it make sense to keep backporting fixes from 5.11.1 given that it will be released next week?

a17r:
Oh, will it indeed? If we can trust upstream there, then we can probably do without.

Pesa:
I'm reasonably sure they want to release `.1` before the summer break. And the last few releases have been relatively on time (±1 week...)

Pesa:
http://lists.qt-project.org/pipermail/releasing/2018-June/004653.html

a17r:
Added disable-xml-catalogs patch from Arfrever for testing.

a17r:
should we adopt the opensuse patch to disable gpu in qtwebengine if nouveau? https://github.com/OpenMandrivaAssociation/qt5-qtwebengine/blob/master/disable-gpu-when-using-nouveau-boo-1005323.diff

okias:
@Pesa time to merge  :+1:  ?

dev-qt/qtwebengine/qtwebengine-5.11.1.ebuild
```diff
@@ -79,6 +79,8 @@ DEPEND="${RDEPEND}
 	pax_kernel? ( sys-apps/elfix )
 "
 
+PATCHES+=( "${FILESDIR}/${P}-libxml2-disable-catalogs.patch" )
```
Pesa: can you reference a bug# either here or in the commit msg?

dev-qt/qtwebengine/files/qtwebengine-5.11.1-ffmpeg4.patch
```diff
@@ -0,0 +1,32 @@
+From a568ded46a678eac8139cb06595819c5ae874177 Mon Sep 17 00:00:00 2001
+From: Xiaohan Wang <xhwang@chromium.org>
+Date: Mon, 29 Jan 2018 21:28:28 +0000
+Subject: [PATCH] media: Increase DecoderBuffer::kPaddingSize to 64
```
Pesa: is this backward compatible with ffmpeg-3?

dev-qt/qtwebengine/qtwebengine-5.11.1.ebuild
Pesa: not sure why github shows this diff as "outdated"...

a17r:
> only the first commit in the series is signed?

must have been caused by rebasing...

okias:
for qtwidgets I get: ERROR: Feature 'gtk3' was enabled, but the pre-condition 'features.glib && libs.gtk3' failed.

I'll try update glib before and lets see

okias:
works fine.

Tested-by: David Heidelberg <david@ixit.cz>

dev-qt/qtwebengine/qtwebengine-5.11.1.ebuild
Pesa: ping on this, the rest LGTM

a17r: thanks, done!

PR #177, opened 2018-06-20
dev-qt/qtwebkit: Add 5.212.0_pre20180120 snapshot

a17r:
Apparently dev-qt/qtwebkit-5.9.1 is showing its age and silently fails
to build on several users' systems, missing required interfaces.

- Switch to cmake buildsystem
- Add USE hyphen, nsplugin, X
- Merge USE webchannel (apparently no revdeps) into USE qml

This snapshot is bumping qtwebkit to the most recent commit by upstream,
undoubtedly also fixing security issues on the way.

Based on work by: Konstantin Tokarev <annulen@yandex.ru>
Tested-by: Manuel Nickschas <sputnick@quassel-irc.org>
Bug: https://bugs.gentoo.org/570534
Bug: https://bugs.gentoo.org/656232
Bug: https://bugs.gentoo.org/624404
Bug: https://bugs.gentoo.org/658018
Package-Manager: Portage-2.3.40, Repoman-2.3.9


dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild
```diff
@@ -0,0 +1,145 @@
...
+	${PYTHON_DEPS}
+	${RUBY_DEPS}
+	dev-lang/perl
+	dev-util/gperf
+	>=sys-devel/bison-2.4.3
+	sys-devel/flex
+	virtual/pkgconfig
+"
+
+CHECKREQS_DISK_BUILD="1G" # Debug build requires much more see bug #417307
```
Pesa: I'm confused, if debug builds require much more, why are we checking for only 1G with `-g`?

This should be bumped to something like 16G according to the referenced bug, although the comments there are pretty old so things could have changed.

dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild
```diff
@@ -0,0 +1,145 @@
...
+
+S="${WORKDIR}/${COMMIT}"
+
+PATCHES=( "${FILESDIR}/${P}-functional.patch" )
+
+pkg_pretend() {
+	if [[ ${MERGE_TYPE} != "binary" ]] ; then
+		if is-flagq "-g*" && ! is-flagq "-g*0" ; then
```
Pesa: can we merged these two `if`s in one, same as in `pkg_setup`

dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild
```diff
@@ -0,0 +1,145 @@
+# Copyright 1999-2018 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=6
+
+COMMIT=72cfbd7664f21fcc0e62b869a6b01bf73eb5e7da
+CMAKE_MAKEFILE_GENERATOR="ninja"
+PYTHON_COMPAT=( python2_7 )
+QT_MIN_VER="5.9.1:5"
```
Pesa: can we move this assignment closer to RDEPEND and friends?

dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild
```diff
@@ -0,0 +1,145 @@
...
+HOMEPAGE="https://www.qt.io/"
+SRC_URI="http://code.qt.io/cgit/qt/${PN}.git/snapshot/${COMMIT}.tar.gz -> ${P}.tar.gz"
+
+if [[ ${PV} != *9999* ]]; then
+	KEYWORDS="~amd64 ~arm ~arm64 ~ppc64 ~x86"
+fi
+
+SLOT=5/5.212
+LICENSE="BSD LGPL-2+"
+IUSE="+geolocation gles2 +gstreamer +hyphen +jit multimedia nsplugin opengl orientation +printsupport qml webp X"
```
Pesa: I'm not sure if I'd enable `geolocation` by default... qtwebengine and previous versions of qtwebkit don't, and we've traditionally been rather minimalistic in this area.

dev-qt/qtwebkit/qtwebkit-5.212.0_pre20180120.ebuild
a17r: the original ebuild was based off a much, much outdated state. this is too low for sure.

a17r:
Should we consider https://bugs.gentoo.org/603880 as well?

a17r:
Okay, with a little bit of research I found this: https://github.com/annulen/webkit/issues/454#issuecomment-283114341

So we can close the issue with the 5.212 bump as well.

a17r:
We should move this to tree together with Qt 5.11.1, please ack when you think it is ready.

Pesa:
https://github.com/gentoo/gentoo/commit/ea2ae777d028e584cd175b2d4081dea826c8f7d1

PR #183, opened 2018-12-10
dev-qt/qtwebengine: Disentangle from Qt version

a17r:
See also:
https://doc.qt.io/qt-5/qtwebengine-platform-notes.html#using-earlier-qt-versions-to-build-qt-webengine

Thanks-to: Jimi Huotari <chiitoo@gentoo.org>
Package-Manager: Portage-2.3.51, Repoman-2.3.11

### DIFF
--- a/dev-qt/qtwebengine/qtwebengine-5.11.9999.ebuild
+++ b/dev-qt/qtwebengine/qtwebengine-5.11.9999.ebuild
@@ -2,6 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=6
+
 PYTHON_COMPAT=( python2_7 )
 inherit multiprocessing pax-utils python-any-r1 qt5-build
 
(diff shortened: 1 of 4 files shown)


dev-qt/qtwebengine/qtwebengine-5.11.9999.ebuild
```diff
@@ -15,17 +16,18 @@ IUSE="alsa bindist designer geolocation +jumbo-build pax_kernel pulseaudio
 	+system-ffmpeg +system-icu widgets"
 REQUIRED_USE="designer? ( widgets )"
 
+QT_PV="5.11.1"
```
Arfrever: Maybe `QT_PV="5.11.1:="` would be better, since in QtWebEngine code there are possibly some `#if` conditionals with versions of other Qt packages...

a17r: probably yes, however it is not yet clear if we can make it work with revdeps anyway. it seems that qtwebengine-installed cmake config insists on same-version qt dependencies still, don't know yet how Fedora and Debian do it.

Pesa:
Conceptually I'm ok with this, but make sure that revdeps can actually be built against a set of qt modules with different version numbers, when I tried (it was a long time ago, 5.6 or even earlier) I had very mixed results and ultimately gave up.

PR #186, opened 2019-01-23
lxqt-base/lxqt-meta: optionally pull in app-arch/lxqt-archiver

Fat-Zer:


Chiitoo:
Apologies for somehow not noticing this until now.

I haven't added 'lxqt-archiver' to the meta yet since it's still in a 'pre-beta' state, but maybe it should be added sometimesoon™ for the live version.

I'll give it some more time and thought.

In any case, thank you!

Chiitoo:
Hrm, I suppose I completely forgot about this at some point... and the flag indeed made its way there some ages past.

In any case, thank you!

PR #187, opened 2019-01-31
qt5-build.eclass: Relocate QT5_DOCDIR to /usr/share/qt5-doc 

a17r:
That's one way out of the QA trouble (and simplifies KDE Frameworks QCH location detection).


a17r:
@pesa switched to using `/usr/share/`**`qt5-doc`** for being more future-proof.

PR #189, opened 2019-05-10
qtlocation: Fix build of 5.13

jktjkt:
Without this patch, the build fails with qmake's error message:
```
Project ERROR: Could not find feature location-labs-plugin.
```
Tested on `5.13.9999`; the `5.9999` (targetting the dev branch upstream) is untested.


kensington:
Thanks!

PR #191, opened 2019-06-07
app-arch/lxqt-archiver: add RDEPEND dev-libs/json-glib

arthurzam:
`lxqt-archiver` depends on `json-glib`, as found from [CMakeLists.txt](https://github.com/lxqt/lxqt-archiver/blob/master/CMakeLists.txt) and from `ldd /usr/bin/lxqt-archiver | grep libjson-glib`.

Package-Manager: Portage-2.3.67, Repoman-2.3.14


Chiitoo:
Thanks!

PR #192, opened 2019-06-18
qt5-build.eclass: EAPI-7 bump

a17r:


a17r:
I've already prepared (and successfully built) Qt 5.12.4 after the change.

PR #193, opened 2019-06-20
Qt 5.12.4 version bump

a17r:


a17r:
let's give it some testing in qt overlay then, and maybe someone comes up with a fix for qtwebengine[geolocation]

PR #195, opened 2019-07-12
dev-python/pyside: 5.12.3 EAPI-7 bump

leycec:
This is the last in a series of pull requests bumping PySide2 & Friends to the "recently" <sup>*wink, wink*</sup> released first stable version of PySide2: [**PySide2 5.12.3**](https://blog.qt.io/blog/2018/12/18/qt-python-5-12-released).

This PR bumps both **`pyside:2`** and **`pyside-tools:2`** – the 500 lb. boa constrictors in the room. Since both require `shiboken:2`, this PR naturally requires [the first PR in this series](https://github.com/gentoo/qt/pull/194). Since that PR has yet to be accepted, this PR is currently just for show. Such is life for the stalwart Gentooer.

*Let's do this yet again!* :boom:

## So, What Exactly Did You Do?

Alotta stuff. The PySide2 codebase has improved more than a little since we authored these ebuilds an ageless aeon ago. Rather than exhaustively enumerate each exhausting detail, here are the highlights:

* **Renamed the `declarative` USE flag to `qml`.** This might be the most contentious change, so it gets top billing. Justifications for this modest change include:
  * **Consistency.** By renaming `declarative` to `qml`, we're then able to trivially forward this flag to all dependencies also declaring this flag (e.g., `charts? ( >=dev-qt/qtcharts-${PV}:5[qml?] )` in `${RDEPEND}`). This improves packaging consistency and system integrity for end users.
  * **Orthogonality.** At least 18 Portage packages prefer `qml` to `declarative`. Only 3 packages still use `declarative` (i.e., `PyQt5`, `designer`, and `qt-docs`), probably for legacy reasons.
  * **Disambiguity.** Since [Qt 5 deprecated the `QtDeclarative` module](https://doc.qt.io/qt-5/qtquick-porting-qt5.html#c-code), the `declarative` USE flag is arguably ambiguous and no longer descriptive. Of course, QML *is* a declarative language. But so are CSS, HTML, SQL, XML, and Lisp. `</shrug>`
[...]

PR #197, opened 2019-07-23
dev-python/shiboken: fix python versioning in cmake targets

leycec:
This is the next in a series of pull requests bumping PySide2 & Friends to the "recently" wink, wink released next stable version of PySide2: **PySide2 5.12.4.**

This PR patches the recently bumped Shiboken2 5.12.4 ebuild to resolve a [severe upstream build issue](https://bugreports.qt.io/browse/PYSIDE-1053) currently blocking #195, the last PR in this series. This PR is thus a mandatory dependency of #195. Once merged, #195 should have no further blockers.

...*should.* :anger: 

# ELI5 or I'm Outta

Currently, Shiboken2 usage is confined to a single Python interpreter – specifically the last Python interpreter against which Shiboken2 was built, which tends to be the highest enabled Python 3 interpreter under Gentoo. This has the obvious adverse effect of breaking Shiboken2 for every *other* enabled Python interpreter, which tends to be Python 2.7 and older enabled Python 3 interpreters under Gentoo.

For voluminous details, see also [**PYSIDE-1053**](https://bugreports.qt.io/browse/PYSIDE-1053).

# What Did You Do About This Horrible Thing?

**One-linered it,** of course! Thanks to the luck of the Canadian Irish, I've uncovered an even *more* efficient means of resolving this on our end than the [awfully convoluted two-step solution detailed at PYSIDE-1053](https://bugreports.qt.io/browse/PYSIDE-1053). That's nice. Here's how it works.

Through the magic of `sed`, we globally replace all *static* references to a specific `libshiboken2*.so` library within the auto-generated `/usr/lib64/cmake/Shiboken-5.12.4/Shiboken2Targets-gentoo.cmake` file (e.g., `/usr/lib64/libshiboken2-python3.6.so.5.12.4`) with *dynamic* references to the current `libshiboken2*.so` library being linked against by PySide2 (e.g., `/usr/lib64/libshiboken2${PYTHON_CONFIG_SUFFIX}.so.5.12.4`).

This works because our PySide2 ebuild will – *once bumped by #195, anyway* – override the `${PYTHON_CONFIG_SUFFIX}` cmake variable to an appropriate string specific to the current Python interpreter being installed against. Wondrous magic follows.

For no good reason, I also accidentally reverted the copying of `/usr/lib64/pkgconfig/shiboken2*.pc` files to mere moving of those files. Either way appears to behave as expected. Being fundamentally lazy, I left it as is. I'd be happy to revert this reversion. Just lemme know!

## I'm Outta

Shout-outs to @Pesa for his perpetual patience during this "process." The promised land is within sight.

### DIFF
--- a/dev-python/shiboken/shiboken-5.12.4-r1.ebuild
+++ b/dev-python/shiboken/shiboken-5.12.4-r1.ebuild
@@ -98,7 +98,19 @@ src_test() {
 src_install() {
 	shiboken_install() {
 		cmake-utils_src_install
+
+		# Uniquify this pkgconfig file to a version-specific basename. Although
+		# each successive call to this function overwrites the original file,
+		# copying rather than moving this file ensures that an original file
+		# targetting the last Python target is still usable by third parties.
 		cp "${ED}/usr/$(get_libdir)"/pkgconfig/${PN}2{,-${EPYTHON}}.pc || die
 	}
 	python_foreach_impl shiboken_install
+
+	# CMakeLists.txt installs a "Shiboken2Targets-gentoo.cmake" file forcing
+	# downstream consumers (e.g., PySide2) to target one "libshiboken2-*.so"
+	# library linked to a single Python interpreter. See also:
+	#     https://bugreports.qt.io/browse/PYSIDE-1053
+	sed -i -e 's~libshiboken2-python[[:digit:]]\+\.[[:digit:]]\+~libshiboken2${PYTHON_CONFIG_SUFFIX}~g' \
+		"${ED}/usr/$(get_libdir)/cmake/Shiboken2-${PV}/Shiboken2Targets-gentoo.cmake" || die
 }


leycec:
**Bumped!** Forgot we needed to force-bump `shiboken-5.12.4.ebuild` to `shiboken-5.12.4-r1.ebuild` to ensure existing Shiboken2 installations receive this critical bugfix. All should be well, now. :sunflower:

Pesa:
I need to study this issue more as I don't fully understand it. Btw, is there any way we can come up with an upstreamable patch?

Pesa:
> For no good reason, I also accidentally reverted the copying of /usr/lib64/pkgconfig/shiboken2*.pc files to mere moving of those files. Either way appears to behave as expected. Being fundamentally lazy, I left it as is. I'd be happy to revert this reversion. Just lemme know!

"either way behaves as expected" I guess refers to pyside, right? have you tested with any other software? (is there *any* other software that uses shib**R**oken??)

I had a reason (maybe not a good one) for the `cp`. Third-party software, as in "not managed by portage", possibly not even installed on the system, but manually built and run. That's a use case we still need to support, think about software developers whose projects use shiboken. Of course pyside works either way, because we're telling the build system where/how to find the renamed files. But 3rd party software doesn't know that we're applying this ugly hack! So by copying the file we will hopefully provide a better out-of-the-box experience, at least for the python target that gets built last (because each target will overwrite the file installed by the previous target).

a17r:
> have you tested with any other software? (is there _any_ other software that uses shib**R**oken??)

`www-client/falkon` would be such a software, but it would have to be `9999` from over at kde overlay as the switches are disabled in Gentoo ebuild repo to avoid automagic deps.

leycec:
> Btw, is there any way we can come up with an upstreamable patch?

**Absolutely.** [The exact changeset required by upstream is documented at PYSIDE-1053]( https://bugreports.qt.io/browse/PYSIDE-1053). If my faulty memory serves, only two lines spread across two Cmake-specific files require patching at installation time. In other words: *super trivial.*

For my part, it's mostly a matter of dwindling free time to devote to open-source volunteerism – as evidenced by this being my first free night in over three weeks. I can only shake my fist at the sky. :cloud:  :fist_left:

leycec:
> have you tested with any other software?

**Nope!** As you surmised, ain't no such software.<sup>1</sup> This is a good thing, though. Shiboken2 has been a dangerously moving target for half a decade now – *and it's still twitching.*

I wouldn't bet my production-ready Python stack on an in-house vendor-locked proprietary C++ parser that gets rewritten from the ground up every two years. Clearly, I'm not the only one.

Of course, the [official Shiboken2 FAQ](https://doc.qt.io/qtforpython/shiboken2/faq.html) suggests they'd *really* like other projects to start leveraging Shiboken2 as a general-purpose C++-to-Python binding generator. There's even a [blog post from a Qt consultancy demonstrating Shiboken2 usage on a toy example](https://blog.basyskom.com/2019/using-shiboken2-to-create-python-bindings-for-a-qt-library). *Somebody* must think it's ready.

That person is not me.

<sup>1. Actually, this isn't quite true. As @a17r helpfully noted, [**falkon**](https://en.wikipedia.org/wiki/Falkon) appears to now optionally require Shiboken2. Ironically, [falkon fails at compile-time on Gentoo systems installing Shiboken2 from this overlay](https://forums.gentoo.org/viewtopic-t-1092364-start-0.html). Falkon is almost certainly at fault here. Due to this, however, we'll need to patch falkon (e.g., by disabling falkon's Shiboken2 support for the moment) *before* merging Shiboken2 into the Portage tree. Since falkon is already written in C++, its dependency on Shiboken2 is presumably just to generate optional Python bindings that no one wants anyway.

> (is there any other software that uses shib**R**oken??)

This bears repeating, people:

> shib**R**oken.

*Oh snap!* You so punny, @Pesa. Thanks to that gut-rumbling chuckle, I feel much better now.

> I had a reason (maybe not a good one) for the `cp`. Third-party software, as in "not managed by portage", possibly not even installed on the system, but manually built and run. That's a use case we still need to support, think about software developers whose projects use shiboken. Of course pyside works either way, because we're telling the build system where/how to find the renamed files. But 3rd party software doesn't know that we're applying this ugly hack! So by copying the file we will hopefully provide a better out-of-the-box experience, at least for the python target that gets built last (because each target will overwrite the file installed by the previous target).

That's... **a fantastic reason,** actually. Thanks for the heady explanation. You're entirely correct, of course. I feel deep shame that I never thought of that. This is why I'm not the repository maintainer.

I'll revert back to the `cp` approach with additional commentary as to why.

leycec:
**Boom.** `mv` reverted back to `cp`, because @Pesa is a light in the darkness and ~~usually~~ always right. :candle:

leycec:
**So.** I come bearing optimistic good news... and horrifying bad news!

The good news is that [Qt engineer Alexandru Croitor](https://bugreports.qt.io/secure/ViewProfile.jspa?name=alexandru.croitor) [promptly replied](https://bugreports.qt.io/browse/PYSIDE-1053?focusedCommentId=469347&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-469347) to [PYSIDE-1053](https://bugreports.qt.io/browse/PYSIDE-1053). *Yay!*

The bad news is that `cmake` (or perhaps upstream's specific usage of `cmake`) appears to offer no trivial hooks or configurability for uniquifying autogenerated `cmake` targets across multiple concurrent Python versions. *Boo!* Specifically, [Alexandru's insightful commentary reads](https://bugreports.qt.io/browse/PYSIDE-1053?focusedCommentId=469347&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-469347):

> While the idea is sound, unfortunately I do not know of any clean way of modifying the autogenerated `Targets` files, bar for using something like `sed`. We'll have to see if there's some other way to do this, maybe generate the `Targets` files multiple times.

The real good news is that this deflated ball is firmly in their court. They openly acknowledge the issue and supposedly intend to eventually resolve it.

The real bad news is I'm allergic to equivocal words like "intend," "supposedly," and "eventually." Let us communally pray as one Gentoo Brotherhood that upstream follows through on this intention.

*So say we all.*

leycec:
**Sure!** KISS for the win. Simplicity is good, so both of those work for me.

That said, Shiboken2 and PySide2 *do* behave as expected after merging this and #195 – multiple Python versions, Python 2.7, warts, and all. Given how much volunteer effort I've ~~squandered~~ invested into this, I'm a bit reticent to go back to the drawing board and fundamentally refactor these ebuilds *yet again.* While dropping Python 2.7 support is (of course) trivial, I'm concerned that downgrading from `python-r1` to `python-single-r1` could prove surprisingly horrifying (i.e., less trivial).

A simple use case for continuing to support multiple Python versions would be [BETSEE](https://gitlab.com/betse/betsee): a PySide2-based multiphysics simulator co-authored by my wife and I. Since all existing Shiboken2 and PySide2 ebuilds currently support multiple Python versions, so do [all existing BETSEE ebuilds](https://github.com/leycec/raiagent/tree/master/sci-biology/betsee). Downgrading Shiboken2 and PySide2 to `python-single-r1` would necessitate downgrading BETSEE and *all* other PySide2-based ebuilds now and forever (e.g., FreeCAD) as well.

Maybe that's not a bad thing? But maybe it is? If we could preserve the flexibility of `python-r1` for the good of future generations, that might be preferable – especially as everything currently works as intended on my end. (*Your mileage may vary, eh?*)

But... yeah. You're the overlay final boss. To [paraphrase the mighty Aku](https://www.youtube.com/watch?v=apnrixZEGq4), "Your rule is law!" Let me know which way you'd like me to run with this. Whatever the outcome, I'll happily see to it. <sup>Well, *mostly* happily.</sup>

leycec:
So.... **it comes to this.**

I don't pretend to understand why this and #195 have come to an abrupt standstill. What I *do* understand is the psychology of previous investment. I've invested far too much unpaid labour on `dev-python/shiboken:2` and `dev-python/pyside:2` to want to see that labour squandered on a vast, impenetrable nothingburger of silence and apathetic shrugs.

All of which is to say: I'll be hosting `dev-python/shiboken:2` and `dev-python/pyside:2` at [my third-party overlay](https://github.com/leycec/raiagent) until further notice. It's obscene that these ebuilds have been working as advertised for several months but unpublished due to... reasons? I have no idea what those reasons might be, but they must be valid – otherwise, they would have been merged by now.

This probably also goes without saying, but I also no longer wish to be a proxy maintainer for these ebuilds. If this overlay is a modest glimpse into the inner sanctums of hallowed Gentoo development, it's little surprise that Arch has absorbed our entire mind- and marketshare. `</collective_shrug>`

As a technical footnote, I should also add that downgrading these ebuilds from `python-r1` to `python-single-r1` **fundamentally does not work.** Sane PySide2 ebuilds need to be usable everywhere that sane PyQt5 ebuilds are usable. This means `dev-python/matplotlib` and `dev-pthyno/QtPy` ebuilds. This means *any* ebuild that currently requires `dev-python/PyQt5`. A `dev-python/pyside:2` ebuild that is *not* globally substitutable for `dev-python/PyQt5` is useless.

*That is all.* Have an excellent Autumn, all. @leycec, out! :door:

PR #199, opened 2019-08-08
qt5-build.eclass: Add split docs handling for >=Qt-5.12.4

a17r:
Copy of https://github.com/gentoo/gentoo/pull/12575 to test with live ebuilds.

Additional tarballs available that I don't know where to add:
```
qtcanvas3d
qtdistancefieldgenerator
qtnfc
qtplatformheaders
qtremoteobjects
tqtc
```

Probably irrelevant: `activeqt`, `qtandroidextras, qtlabsplatform`, `qtlabscalendar`, `qtmacextras`, `qtwinextras`

On the other hand, qt-docs could also be repurposed to contain these remaining docs.


Pesa:
> Additional tarballs available that I don't know where to add:
> 
> qtcanvas3d

This is deprecated upstream. Do not add it.

> qtdistancefieldgenerator

This was added in 5.12 but I can't find the repo/module it belongs to.

> qtnfc

Not packaged in gentoo yet.

> qtplatformheaders

This goes in `qtgui`

> qtremoteobjects

Not packaged in gentoo yet.

> tqtc

No idea what this is.

> Probably irrelevant: `activeqt`, `qtandroidextras, qtlabsplatform`, `qtlabscalendar`, `qtmacextras`, `qtwinextras`

Yeah I'd ignore those.

a17r:
So, while testing we came to the conclusion this would lead to re-manifesting live ebuilds on version bumps. Which is certainlypossible, but adding complexity. Also, we are adding `doc` USE flags to packages, some quite heavy on cpu time, just to add ready-made documentation tarballs to them. Even if it is elegant to get the necessary documentation fitted per package, I don't think it is worth the code rebuilds it is causing when USE doc is flipped, and we should defer this until portage is capable of optional runtime deps.

PR #204, opened 2020-01-29
eclass: do not pass "-nomake tests" to configure when we want Qt tests

DerDakon:
Moving [gentoo#14222](https://github.com/gentoo/gentoo/pull/14222) to the right place.


Pesa:
what's the rationale for these changes?

DerDakon:
Trying to sort out strange things happening in the Gentoo build system while trying to report things upstream. They kept asking me how it can happen that I have tests when passing -no-make tests and things like that.

PR #207, opened 2020-02-09
dev-qt/qtnetwork: [5.15] add patch for LibreSSL support

steils:
Bug: https://bugs.gentoo.org/562050
Package-Manager: Portage-2.3.87, Repoman-2.3.20
Signed-off-by: Stefan Strogin <steils@gentoo.org>

### DIFF
--- a/dev-qt/qtnetwork/files/qtnetwork-5.15.9999-libressl.patch
+++ b/dev-qt/qtnetwork/files/qtnetwork-5.15.9999-libressl.patch
@@ -1,0 +1,340 @@
+From 4774fcd31a49f6f193bf10990601ad494fab2013 Mon Sep 17 00:00:00 2001
+From: Stefan Strogin <steils@gentoo.org>
+Date: Wed, 5 Feb 2020 03:49:35 +0200
+Subject: [PATCH] QSslSocket - add LibreSSL support
+
+Upstream-Status: Inappropriate
+[Upstream is not willing to accept any patches for LibreSSL support]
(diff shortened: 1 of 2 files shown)


steils:
@Pesa ?

steils:
Created a PR for main Gentoo repository, as qt-5.15.0 was moved there: https://github.com/gentoo/gentoo/pull/14494

This is still actual. Please look at it.

a17r:
for 5.15.9999 it makes sense as releases are made out of it, 5.9999 will be mostly broken by now I think

Pesa:
Right. 5.9999 should probably not even exist anymore...

steils:
I think applying it on 5.15.9999 (and 5.16.9999 when the time comes) will be useful to warn me if there are any incompatibilities with the upcoming Qt release, so I could update the patch in time.

Pesa:
makes sense (note: there won't be a 5.16, the next version is 6.0)

steils:
Yay! It already fails to apply :)
I'll fix it.

steils:
I have updated the patch. The incompatibility was caused by this patch: https://code.qt.io/cgit/qt/qtbase.git/commit/?h=5.15&id=1941ec43ced63a1f82ca9ddad1c71d5179c33b2d
they pushed it into 5.15 branch.

Also while my patch in Gentoo was under review, LibreSSL added and enabled by default TLS 1.3 support:) So I think I'll be able to make the patch smaller thanks to that. I'll begin working on it in the next days.

PR #209, opened 2020-03-17
dev-qt/qtwayland: Drop USE libinput and xcomposite

a17r:


Pesa:
I'm not sure if I agree with dropping `USE=libinput` from qtgui. If qtwayland needs libinput why cannot it depend on `qtgui[libinput]`?

Also, forcing `xcomposite` in qtwayland means hard dependencies on X11 stuff... I think the idea behind that flag was avoiding those dependencies. Is there any way to fix it instead of dropping it?

a17r:
> I'm not sure if I agree with dropping `USE=libinput` from qtgui. If qtwayland needs libinput why cannot it depend on `qtgui[libinput]`?

Right, not sure why I didn't see that yesterday night.

> Also, forcing `xcomposite` in qtwayland means hard dependencies on X11 stuff... I think the idea behind that flag was avoiding those dependencies. Is there any way to fix it instead of dropping it?

Well, so far it was using automagic on most systems and depended on X11 stuff anyway. I'm not sure of the implications on revdeps if now really making it conditional.

Pesa:
> > Also, forcing `xcomposite` in qtwayland means hard dependencies on X11 stuff... I think the idea behind that flag was avoiding those dependencies. Is there any way to fix it instead of dropping it?
> 
> Well, so far it was using automagic on most systems and depended on X11 stuff anyway. I'm not sure of the implications on revdeps if now really making it conditional.

Ok, let's make it mandatory for now. We can revisit for Qt 6. I suppose it's too early for a "pure wayland" system anyway.

a17r:
> Ok, let's make it mandatory for now. We can revisit for Qt 6. I suppose it's too early for a "pure wayland" system anyway.

After some qmake args provided by @Chiitoo going to keep the option and renaming to X (as requested by x11 proj). It will be a minority setup anyway, and what little testing it gets we can wait and see.

PR #210, opened 2020-03-21
dev-qt/qtwayland: USE X requires !dev-qt/qtgui[gles2]

a17r:
So the first results are in wrt fixing the dev-qt/qtwayland[X] automagic by using the appropriate build switches: https://bugs.gentoo.org/713362

Please note that I think the dependency is suboptimal, even if necessary.

This affects users who globally enable gles2 for whatever reasons and don't use desktop profile for whatever reasons, where this is masked for Qt packages.

The best way to tackle this imo is to finally rename the gles2 flag, see also: https://archives.gentoo.org/gentoo-dev/message/ab302ac2ab641b1e82acad7d32308266

For `gles2-only` there is already usage found in net-libs/webkit-gtk and x11-libs/cairo.


a17r:
oh, err yes. doesn't change the unfortunate link between X and gles2 though.

PR #211, opened 2020-03-21
dev-qt/*: Rename USE=gles2 to USE=gles2-only

a17r:
See also:
https://bugs.gentoo.org/627758
https://archives.gentoo.org/gentoo-dev/message/ab302ac2ab641b1e82acad7d32308266
https://github.com/gentoo/gentoo/pull/13742
https://github.com/gentoo/qt/pull/210

### DIFF
--- a/dev-qt/qt3d/metadata.xml
+++ b/dev-qt/qt3d/metadata.xml
@@ -6,9 +6,7 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
-		<flag name="gamepad">Add support for gamepad hardware via
-			<pkg>dev-qt/qtgamepad</pkg></flag>
-		<flag name="gles2">Use GLES 2.0 or later instead of full OpenGL</flag>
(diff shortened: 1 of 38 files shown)


a17r:
there was not a lot of participation, so we can assume silent agreement? at least no one from Qt proj took part to suggest something else in the old bug from 2018 or the ML, and since then x11 and gnome proj have just silently started using this flag.

In webkit-gtk, it basically means the same: `Use GLES 2.0 or later instead of full OpenGL`
Only cairo at least pretends to limit to v2: `Build the OpenGL ES 2 backend`

lanodan:
Might be outside the scope of this overlay but you should also change it for revdeps, for example kdenlive (with `IUSE="… gles2 …"`) depends on `>=dev-qt/qtgui-${QTMIN}:5[gles2=]`, which would break otherwise.

a17r:
yes that's out of scope here, we can only try to coordinate as good as possible. there is also kde overlay and changes always go first into overlays, then to tree.

lanodan:
I guess I could do another PR for stuff outside of Qt/KDE then? I think I should keep the big PR as-is for the main tree at least as a base. Should I or you do the PR to KDE?

PR #212, opened 2020-04-01
dev-qt: Add more slot operators

a17r:
In the past we had many issues related to inconsistent Qt packages. Quite simply it is possible for users to end up with half Qt 5.13.x, half 5.14.x installed. The addition of finer-grained subslots because of frequent private ABI changes between patch releases, and utilisation in affected reverse-dependencies had an unintended side-effect: They are being queued up earlier for emerge, way before all of dev-qt/* has been upgraded, making failed builds way more likely.

This package is intended to help the package manager figure out a better order.

The straw that broke the camel's back: https://gist.github.com/a17r/fc35a8582ab8ed7baa23c2cde85cf1cc

Emerge queue after the suggested change:
https://gist.github.com/a17r/dec177c618b0e28d03db668e574098fc

### DIFF
--- a/dev-qt/designer/designer-5.15.9999.ebuild
+++ b/dev-qt/designer/designer-5.15.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -14,8 +14,8 @@ fi
 IUSE="declarative webkit"
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
-	~dev-qt/qtgui-${PV}
+	~dev-qt/qtcore-${PV}:5=
+	~dev-qt/qtgui-${PV}:5=
 	~dev-qt/qtnetwork-${PV}
 	~dev-qt/qtprintsupport-${PV}
 	~dev-qt/qtwidgets-${PV}
--- a/dev-qt/linguist/linguist-5.14.9999.ebuild
+++ b/dev-qt/linguist/linguist-5.14.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -15,8 +15,8 @@ IUSE=""
 
 DEPEND="
 	~dev-qt/designer-${PV}
-	~dev-qt/qtcore-${PV}
-	~dev-qt/qtgui-${PV}
+	~dev-qt/qtcore-${PV}:5=
+	~dev-qt/qtgui-${PV}:5=
 	~dev-qt/qtprintsupport-${PV}
 	~dev-qt/qtwidgets-${PV}
 	~dev-qt/qtxml-${PV}
--- a/dev-qt/pixeltool/pixeltool-5.15.9999.ebuild
+++ b/dev-qt/pixeltool/pixeltool-5.15.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -14,8 +14,8 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
-	~dev-qt/qtgui-${PV}
+	~dev-qt/qtcore-${PV}:5=
+	~dev-qt/qtgui-${PV}:5=
 	~dev-qt/qtwidgets-${PV}
 "
 RDEPEND="${DEPEND}"
--- a/dev-qt/qtconcurrent/qtconcurrent-5.14.9999.ebuild
+++ b/dev-qt/qtconcurrent/qtconcurrent-5.14.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2020 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:5=
 "
 RDEPEND="${DEPEND}"
 
--- a/dev-qt/qtcore/qtcore-5.14.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.14.9999.ebuild
@@ -2,10 +2,12 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
+
 QT5_MODULE="qtbase"
 inherit linux-info qt5-build
 
 DESCRIPTION="Cross-platform application development framework"
+SLOT=5/$(ver_cut 1-3)
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~sparc ~x86"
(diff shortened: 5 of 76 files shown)


a17r:
@Pesa if possible I'd like to add it for 5.14.2 already, if you don't object.

Pesa:
TBH I'm not sure I like this, it looks like a giant hack. We should fix our dependencies instead. If qt module A depends on B-private, we need to add a subslot dep from B to A. This is probably common for qtbase modules. Wouldn't that solve the problem?

Pesa:
(and probably more modules need to use the full `5.x.y` as their SLOT)

Chiitoo:
Wasn't adding subslot deps what caused this issue in the first place though?

I wonder if there's a Portage bug actually that could be fixed for this.

Chiitoo:
Meh, half-asleep reading/commenting no good doing.

Can probably disregard my previous thoughts here.

Pesa:
My understanding of @a17r's gist is that `libQt5PrintSupport.so` is being used during linking of another revdep (breeze-settings5?) *before* it was rebuilt/upgraded. And `libQt5PrintSupport.so` is broken because qtgui broke its private ABI, but qtprintsupport doesn't have a sublot dependency so it wasn't scheduled to be rebuilt/upgraded earlier by portage.

I'm assuming (and hoping) that if multiple revdeps are to be rebuilt due to subslot changes, portage will still respect the relative dependencies (including transitive ones) among them while building the depgraph.

a17r:
Yes, adding slot operators on qtcore dependency to all qtbase packages would be reasonable, it would solve part of the inconsistent-Qt-installation problem (for everything except outside of qtbase). That would require an additional finer-grained subslot on `dev-qt/qtcore`. In recent times though I got an impression that Portage's dependency graph ordering is not always optimal - maybe undeservedly so, because of https://bugs.gentoo.org/703306 where Qt broke its CMake backwards compatibility.

a17r:
Results are in for the subslot aided upgrade, non-dev-qt packages are queued much later now:

https://gist.github.com/a17r/db28ad83169dbd6aa610644492f43509

dev-qt/qtconcurrent/qtconcurrent-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
```
Pesa, line 17: is this equivalent to `${SLOT%/*}`? or we could even hardcode `:5=` I guess

dev-qt/pixeltool/pixeltool-5.15.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
 	~dev-qt/qtgui-${PV}
```
Pesa, line 18: pixeltool has a private dependency on qtgui as well

dev-qt/designer/designer-5.15.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE="declarative webkit"
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
 	~dev-qt/qtgui-${PV}
```
Pesa, line 18: designer has a private dependency on qtgui as well

dev-qt/linguist/linguist-5.14.9999.ebuild
```diff
@@ -15,7 +15,7 @@ IUSE=""
 
 DEPEND="
 	~dev-qt/designer-${PV}
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
 	~dev-qt/qtgui-${PV}
```
Pesa, line 19: linguist has a private dependency on qtgui as well

dev-qt/qdbus/qdbus-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
```
Pesa, line 17: I don't think qdbus needs this

dev-qt/qdbusviewer/qdbusviewer-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
```
Pesa, line 17: I don't think qdbusviewer needs this

dev-qt/qtdiag/qtdiag-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE="+network +widgets"
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
 	~dev-qt/qtgui-${PV}
```
Pesa, line 18: qtdiag has a private dependency on qtgui as well

dev-qt/qtpaths/qtpaths-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
```
Pesa, line 17: I don't think this is needed for qtpaths

dev-qt/qtplugininfo/qtplugininfo-5.14.9999.ebuild
```diff
@@ -14,7 +14,7 @@ fi
 IUSE=""
 
 DEPEND="
-	~dev-qt/qtcore-${PV}
+	~dev-qt/qtcore-${PV}:${SLOT/\/*}=
```
Pesa, line 17: ditto for qtplugininfo

dev-qt/qtcore/qtcore-5.14.9999.ebuild
```diff
@@ -2,10 +2,12 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
+
 QT5_MODULE="qtbase"
 inherit qt5-build
 
 DESCRIPTION="Cross-platform application development framework"
+SLOT=5/${PV}
```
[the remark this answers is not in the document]
a17r: thanks for noticing, fixed dev-qt/qtgui right now and will change this as well.

dev-qt/qtconcurrent/qtconcurrent-5.14.9999.ebuild
a17r: it is, in fact, yes.

I was actually wondering why the dependencies stopped short of the slot def already.

In the end, when these ebuilds are going to be ported to Qt6 another run to /dev-qt/s/5/6/ won't matter so I'll just convert to `:5=`.

Pesa: > I was actually wondering why the dependencies stopped short of the slot def already.

No reason I guess. A `:5` SLOT dep is probably redundant in a `~${PV}` style dep. I guess that was the reason.

> In the end, when these ebuilds are going to be ported to Qt6 another run to /dev-qt/s/5/6/ won't matter so I'll just convert to `:5=`.

Sure, I'm fine either way.

Pesa:
Thanks

a17r:
Looks like another case of Portage ordering kde-frameworks/ package too early while dev-qt/* category is not fully upgraded.

https://bugs.gentoo.org/749777

PR #213, opened 2020-04-04
qt5-build.eclass: Drop obsolete in-source build workaround for >=5.14.2

a17r:
Fixed upstream in 5.11.1.

See also: https://bugreports.qt.io/browse/QTBUG-37417
Upstream commit 67aa365d41ebfe082b4efcfd725e4d5f08be678c

Bug: https://bugs.gentoo.org/497312
Bug: https://bugs.gentoo.org/676948

I think at least dev-qt/qdoc will need to be adapted as well for that change.

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -104,10 +104,11 @@ EGIT_REPO_URI=(
 # @OUTPUT_VARIABLE
 # @DESCRIPTION:
 # Build directory for out-of-source builds.
-case ${QT5_BUILD_TYPE} in
-	live)    : ${QT5_BUILD_DIR:=${S}_build} ;;
-	release) : ${QT5_BUILD_DIR:=${S}} ;; # workaround for bug 497312
-esac
+if ver_test ${PV} -lt 5.14.2; then
+	: ${QT5_BUILD_DIR:=${S}} # workaround for bug 497312
+else
+	: ${QT5_BUILD_DIR:=${S}_build}
+fi
 
 IUSE="debug test"
 
@@ -658,7 +659,12 @@ qt5_base_configure() {
 
 	# a forwarding header is no longer created since 5.8, causing the system
 	# config to always be used. bug 599636
-	cp src/corelib/global/qconfig.h include/QtCore/ || die
+	# ${S}/include does not exist in live sources
+	local basedir="${S}/"
+	if ver_test ${PV} -lt 5.14.2 || [[ ${QT5_BUILD_TYPE} == live ]]; then
+		basedir=""
+	fi
+	cp src/corelib/global/qconfig.h "${basedir}"include/QtCore/ || die
 
 	popd >/dev/null || die
 
--- a/dev-qt/qdoc/qdoc-5.14.9999.ebuild
+++ b/dev-qt/qdoc/qdoc-5.14.9999.ebuild
@@ -36,5 +36,6 @@ src_configure() {
 	# run in the root directory. bug 676948; same fix as bug 633776
 	mkdir -p "${QT5_BUILD_DIR}"/src/qdoc || die
 	qt5_qmake "${QT5_BUILD_DIR}"
+	cp src/qdoc/qtqdoc-config.pri "${QT5_BUILD_DIR}"/src/qdoc || die
 	qt5-build_src_configure
 }


Pesa:
oh yeah, I forgot about this...

eclass/qt5-build.eclass
```diff
@@ -657,7 +658,11 @@ qt5_base_configure() {
 
 	# a forwarding header is no longer created since 5.8, causing the system
 	# config to always be used. bug 599636
-	cp src/corelib/global/qconfig.h include/QtCore/ || die
+	local basedir="${S}/"
+	if ver_test ${PV} -lt 5.14.2 || [[ ${QT5_BUILD_TYPE} == live ]]; then
```
Pesa, line 662: live versions don't need `$S`? or are you just trying to preserve the current behavior?

eclass/qt5-build.eclass
```diff
@@ -657,7 +658,11 @@ qt5_base_configure() {
...
+		basedir=""
+	fi
+	cp src/corelib/global/qconfig.h "${basedir}"include/QtCore/ || die
```
Pesa, line 665: we shouldn't touch `$S` if we're doing an out-of-source build... OTOH it doesn't really matter at the moment. Does it really not work without `$S`?

eclass/qt5-build.eclass
a17r: that file path does not exist in build dir if that is not $S, and upstream has some `if .git/ exists` style conditional in build system unfortunately...

eclass/qt5-build.eclass
a17r: since live and release behaviour differ, the other option would have been creaiting a faux .git directory in $S. but I don't know what other changes that might bring.

Pesa: no I wouldn't do that. This whole thing is a giant hack anyway.

eclass/qt5-build.eclass
Pesa: I'm not sure I follow. Are you saying `include/QtCore` does not exist in `QT5_BUILD_DIR` in the *release* case? Maybe we should create the directory hierarchy then? Maybe this workaround is obsolete?

Arfrever: `${S}/include` directory does not exist for live ebuilds (see [upstream repository](https://code.qt.io/cgit/qt/qtbase.git/tree/?h=5.15)), but exists for release ebuilds.

eclass/qt5-build.eclass
Arfrever: See `.git` checks:
https://code.qt.io/cgit/qt/qtbase.git/tree/configure?h=5.15&id=6387138a7991b4588639dc48847f175b5afaff84#n692
https://code.qt.io/cgit/qt/qtbase.git/tree/configure?h=5.15&id=6387138a7991b4588639dc48847f175b5afaff84#n752

eclass/qt5-build.eclass
Chiitoo: I tested these changes without this bottom part, and still got

```
src/qdoc/clangcodeparser.cpp:58:10: fatal error: clang-c/Index.h: No such file or directory
   58 | #include <clang-c/Index.h>
      |          ^~~~~~~~~~~~~~~~~
```

Chiitoo: (That is from qdoc-5.15.0_beta3 and 5.15.9999.)

Chiitoo: Aside from that, at least the 9999 versions (of Qt packages I normally have installed) seem to work fine still.

a17r: Qt 5.14.2 also built fine without this `cp` line. Only keeping it for <5.14.2 then.

Chiitoo: I made a little ooopsies.  I didn't have clang installed (was using 'ebuild' to test 'qdoc' in particular)... but shhh, don't tell anyone.  :V

No wonder I couldn't make 'qdoc' work, no matter what I did.  Regardless, it still doesn't work as is.

Either setting ```QT5_BUILD_DIR="${S}"```, or the following is required (or something completely different):

``` bash
src_configure() {
	mkdir -p "${QT5_BUILD_DIR}"/src/qdoc || die
	qt5_qmake "${QT5_BUILD_DIR}"
	cp "${S}"/src/qdoc/qtqdoc-config.pri "${QT5_BUILD_DIR}"/src/qdoc || die
	qt5-build_src_configure
}
```

Chiitoo: Also, due to upstream commit 50ef2ac2 [1], a similar workaround is required for most other qttools as well, but it can differ a bit from module to module (or at least between 'qdoc' and the others), so I guess we can't do it in the eclass.  Not that it feels like a real fix still...

1. https://code.qt.io/cgit/qt/qttools.git/commit/?h=5.15&id=50ef2ac2

a17r: I confirm this works now for qdoc 5.14.2 and 5.14.9999, thanks. Maybe you can take care of 5.15+ ebuilds when this was ACKed @Chiitoo

a17r: Although I hope we don't run into issues like https://bugs.gentoo.org/498894 again, in case that was related.

eclass/qt5-build.eclass
Pesa: @Arfrever what are you trying to say exactly?

eclass/qt5-build.eclass
Pesa: yeah that's the problem, I don't think you can simply say this isn't needed based on a successful build, unless you can replicate exactly the conditions that triggered the bug. IIRC this was a case of the build system using the (old) installed copy of some header instead of the one being built.

a17r: should we just keep the previous workaround then, even if it touches `$S`?

a17r: I'd like to merge this today so it is clear to @Chiitoo in which way to fix the broken >=5.15 qttools subpackages

It is still good to go back to out-of-source build even if it is not perfect with that workaround.

PR #214, opened 2020-04-08
dev-qt/qtcore: enable renameat2 and getentropy conditionally

heroxbd:
Add USE=old-kernel to switch off the new syscalls.

renameat2() and getentropy() are not available until linux-3.17.

Bug: https://bugs.gentoo.org/669994
Package-Manager: Portage-2.3.88, Repoman-2.3.18
Signed-off-by: Benda Xu <heroxbd@gentoo.org>

### DIFF
--- a/dev-qt/qtcore/metadata.xml
+++ b/dev-qt/qtcore/metadata.xml
@@ -6,6 +6,7 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
+		<flag name="old-kernel">disable new syscalls for Linux kernel >= 3.17</flag>
 		<flag name="systemd">Enable native journald logging support</flag>
 	</use>
 	<upstream>
--- a/dev-qt/qtcore/qtcore-5.14.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.14.9999.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 QT5_MODULE="qtbase"
-inherit qt5-build
+inherit linux-info qt5-build
 
 DESCRIPTION="Cross-platform application development framework"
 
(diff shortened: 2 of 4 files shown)


dev-qt/qtcore/metadata.xml
```diff
@@ -6,6 +6,7 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
+		<flag name="old-kernel">disable new syscalls for Linux kernel >= 3.17</flag>
```
Pesa, line 9: description is confusing, it sounds like the flag should be enabled for kernels >= 3.17, while the opposite is true.

Also, start with a capital letter.

dev-qt/qtcore/qtcore-5.14.9999.ebuild
```diff
@@ -44,6 +44,16 @@ QT5_GENTOO_PRIVATE_CONFIG=(
 
 PATCHES=( "${FILESDIR}/${PN}-5.14.1-cmake-macro-backward-compat.patch" ) # bug 703306
 
+pkg_pretend() {
+	if use kernel_linux; then
```
Pesa, line 48: `use kernel_linux || return` to reduce indentation in the remainder of the function

dev-qt/qtcore/qtcore-5.14.9999.ebuild
```diff
@@ -3,15 +3,15 @@
...
+inherit linux-info qt5-build
 
 DESCRIPTION="Cross-platform application development framework"
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~sparc ~x86"
 fi
 
-IUSE="icu systemd"
+IUSE="icu old-kernel systemd"
```
Pesa, line 14: do we have to add `kernel_linux` here? I forgot how this works... 🙈

dev-qt/qtcore/metadata.xml
a17r: How about `Disable syscalls not available on Linux kernels older than 3.17`

dev-qt/qtcore/qtcore-5.14.9999.ebuild
a17r: It is already provided by `linux-info.eclass`

dev-qt/qtcore/metadata.xml
Pesa: sounds good to me

PR #215, opened 2020-04-08
dev-qt/qtwebengine: ninja options and ICU safeguard

a17r:
Still to be test-built.


dev-qt/qtwebengine/qtwebengine-5.14.9999.ebuild
```diff
@@ -92,6 +92,19 @@ src_prepare() {
 	# bug 620444 - ensure local headers are used
 	find "${S}" -type f -name "*.pr[fio]" | xargs sed -i -e 's|INCLUDEPATH += |&$$QTWEBENGINE_ROOT/include |' || die
 
+	if use system-icu; then
+		# Sanity check to ensure that bundled copy of ICU is not used.
+		# Whole src/3rdparty/chromium/third_party/icu directory cannot be deleted because
+		# src/3rdparty/chromium/third_party/icu/BUILD.gn is used by build system.
```
Pesa, line 98: so can we delete everything inside that dir *except* `BUILD.gn`?

a17r: hm, I guess so, would you prefer that? I haven't started test build anyway yet.

Pesa: yes please, looks like it would be slightly simpler (no `for` loop)

a17r: unfortunately not so easy, there are references from other parts of the chromium build system to subdirs below icu, so who knows how many files will be expected to simply be there...

a17r:
Building 5.14.2 finished successfully with the current state of PR.

dev-qt/qtwebengine/qtwebengine-5.14.9999.ebuild
Pesa: hmm, fine, but the loop should be rewritten like this:

```
while read file; do
    echo ...
done < <(find ... -type f -name ... 2>/dev/null)
```

PR #217, opened 2020-04-10
dev-qt/qtcore: renameat2 is glibc specific

a17r:


Pesa:
Right... this is error-prone. How about we force `-no-foo` when `old-kernel` is enabled, and do nothing otherwise (letting qt's configure do the detection as before)?

anarchpenguin:
This is misleading for musl users. You are asking them to say the kernel is no supported even tho they are on much newer kernel.

a17r:
After this change, 5.14.2 should work the same way as 5.14.1 and before did, by not specifically requesting `-feature-renameat2` but letting Qt's detection work it out. Why would musl users even look at qtcore in that case?

PR #218, opened 2020-04-11
dev-qt: remove argument from 'qt5_qmake'

Chiitoo:
Does this look right?

### DIFF
--- a/dev-qt/designer/designer-5.15.0_beta3.ebuild
+++ b/dev-qt/designer/designer-5.15.0_beta3.ebuild
@@ -44,7 +44,7 @@ src_configure() {
 	# run in the root directory.
 	# Related bugs: 633776, 676948, and 716514.
 	mkdir -p "${QT5_BUILD_DIR}" || die
-	qt5_qmake "${QT_BUILD_DIR}"
+	qt5_qmake
 	cp "${S}"/qttools-config.pri "${QT5_BUILD_DIR}" || die
 	qt5-build_src_configure
 }
--- a/dev-qt/designer/designer-5.15.9999.ebuild
+++ b/dev-qt/designer/designer-5.15.9999.ebuild
@@ -44,7 +44,7 @@ src_configure() {
 	# run in the root directory.
 	# Related bugs: 633776, 676948, and 716514.
 	mkdir -p "${QT5_BUILD_DIR}" || die
-	qt5_qmake "${QT_BUILD_DIR}"
+	qt5_qmake
 	cp "${S}"/qttools-config.pri "${QT5_BUILD_DIR}" || die
 	qt5-build_src_configure
 }
--- a/dev-qt/linguist-tools/linguist-tools-5.15.0_beta3.ebuild
+++ b/dev-qt/linguist-tools/linguist-tools-5.15.0_beta3.ebuild
@@ -39,7 +39,7 @@ src_configure() {
 	# run in the root directory.
 	# Related bugs: 633776, 676948, and 716514.
 	mkdir -p "${QT5_BUILD_DIR}" || die
-	qt5_qmake "${QT_BUILD_DIR}"
+	qt5_qmake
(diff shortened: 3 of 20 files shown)


dev-qt/designer/designer-5.15.0_beta3.ebuild
```diff
@@ -44,7 +44,7 @@ src_configure() {
 	# run in the root directory.
 	# Related bugs: 633776, 676948, and 716514.
 	mkdir -p "${QT5_BUILD_DIR}" || die
-	qt5_qmake "${QT_BUILD_DIR}"
+	qt5_qmake
```
Pesa, line 47: indeed, arguments to `qt5_qmake` are passed via the `myqmakeargs` bash array.

is the intention here to run `qt5_qmake` in `$S`?

Chiitoo: I believe so.  Things do "just work", and I had originally copied one of the ebuilds that had it like so: https://bugs.gentoo.org/633776#c6

Then it somehow snuck into one of the recent commits, where it was supposed to be 'QT5_BUILD_DIR', and then I noticed it doesn't matter what it is... which I actually seem to have first noticed back here: https://bugs.gentoo.org/676948#c1

Never looked into it after that, until now.

Pesa: so it's not just about *root* vs *subdir* as the comment above seems to imply, but also `S` vs `QT5_BUILD_DIR`, i.e. in-source vs out-of-source. Can you clarify the comment please? (and TBH, this is another huge hack, I wonder what's the point of doing out-of-source if we still need to do all these hacks that touch `S`...)

Chiitoo:
While looking for a better way to do this, or rather avoid doing it completely, I figured that for most of the qttools we could just do something like for example for qdoc:

```
src_configure() {
	local myqmakeargs=(
		--
		-no-feature-assistant
		-no-feature-designer
		-no-feature-distancefieldgenerator
		-no-feature-kmap2qmap
		-no-feature-linguist
		-no-feature-macdeployqt
		-no-feature-makeqpf
		-no-feature-pixeltool
		-no-feature-qdbus
		-feature-qdoc
		-no-feature-qev
		-no-feature-qtattributionsscanner
		-no-feature-qtdiag
		-no-feature-qtpaths
		-no-feature-qtplugininfo
		-no-feature-windeployqt
		-no-feature-winrtrunner
	)

	qt5-build_src_configure
}
```

This would include removing the 'QT5_TARGET_SUBDIRS', and as such, 'qt5_qmake' runs in the root directory of the sources, and the file that's copied in this workaround gets generated in the out-of-source build directory just as it should be.

There are some that would require some more thinking, such as assistant, help, and linguist at the very least, but so far this seems to work like we'd want it to.

a17r:
> ``` src_configure() {
> 	local myqmakeargs=(
> 		--
> 		-no-feature-assistant
> 		-no-feature-designer
> 		-no-feature-distancefieldgenerator
> 		-no-feature-kmap2qmap
> 		-no-feature-linguist
> 		-no-feature-macdeployqt
> 		-no-feature-makeqpf
> 		-no-feature-pixeltool
> 		-no-feature-qdbus
> 		-feature-qdoc
> 		-no-feature-qev
> 		-no-feature-qtattributionsscanner
> 		-no-feature-qtdiag
> 		-no-feature-qtpaths
> 		-no-feature-qtplugininfo
> 		-no-feature-windeployqt
> 		-no-feature-winrtrunner
> 	)
> 
> 	qt5-build_src_configure
> }
> ```
> 
> This would include removing the 'QT5_TARGET_SUBDIRS', and as such, 'qt5_qmake' runs in the root directory of the sources, and the file that's copied in this workaround gets generated in the out-of-source build directory just as it should be.

I like that solution a lot, using switches the build system provides. And this could be moved into a qttools conditional block in the eclass, disabling all "featuress" except ${PN}.

PR #221, opened 2020-04-27
qt5-build.eclass: Introduce qt5_tools_configure

a17r:
It seems to work in initial build with 5.15.0_beta4.

### DIFF
--- a/dev-qt/designer/designer-5.15.0_beta4.ebuild
+++ b/dev-qt/designer/designer-5.15.0_beta4.ebuild
@@ -25,10 +25,6 @@ DEPEND="
 "
 RDEPEND="${DEPEND}"
 
-QT5_TARGET_SUBDIRS=(
-	src/designer
-)
-
 src_prepare() {
 	qt_use_disable_mod declarative quickwidgets \
 		src/designer/src/plugins/plugins.pro
@@ -37,16 +33,6 @@ src_prepare() {
 		src/designer/src/plugins/plugins.pro
 
 	qt5-build_src_prepare
-}
-
-src_configure() {
-	# Most of qttools require files that are only generated when qmake is
-	# run in the root directory.
-	# Related bugs: 633776, 676948, and 716514.
-	mkdir -p "${QT5_BUILD_DIR}" || die
-	qt5_qmake "${QT_BUILD_DIR}"
-	cp "${S}"/qttools-config.pri "${QT5_BUILD_DIR}" || die
-	qt5-build_src_configure
 }
 
 src_install() {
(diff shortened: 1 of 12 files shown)


dev-qt/qdbus/qdbus-5.15.0_beta4.ebuild
```diff
@@ -19,7 +19,3 @@ DEPEND="
 	~dev-qt/qtxml-${PV}
 "
 RDEPEND="${DEPEND}"
-
-QT5_TARGET_SUBDIRS=(
-	src/qdbus/qdbus
-)
```
Pesa, line 25: removing this won't end up building both qdbus and qdbusviewer?

eclass/qt5-build.eclass
```diff
@@ -670,6 +673,33 @@ qt5_base_configure() {
...
+		-no-feature-windeployqt
+		-no-feature-winrtrunner
+	)
+
+	for x in assistant designer linguist pixeltool qdbus qdoc qtdiag qtpaths qtplugininfo; do
+		[[ ${PN} = ${x} ]] || qmakeargs+=( -no-feature-${x} )
+	done
+
+	# allow the ebuild to override what we set here
+	myqmakeargs=( "${qmakeargs[@]}" "${myqmakeargs[@]}" )
```
Pesa, line 700: maybe use another array (internal to the eclass) instead of hijacking `myqmakeargs`?

eclass/qt5-build.eclass
```diff
@@ -670,6 +673,33 @@ qt5_base_configure() {
...
+qt5_tools_configure() {
+	# configure arguments
+	local qmakeargs=(
+		--
+		# not packaged in Gentoo
+		-no-feature-distancefieldgenerator
+		-no-feature-kmap2qmap
+		-no-feature-macdeployqt
+		-no-feature-makeqpf
+		-no-feature-qev
```
Pesa, line 689: (off topic) we should probably package this at some point

eclass/qt5-build.eclass
a17r: hm, that means adding it to `qt5_qmake` just for qttools? not sure it is worth it, can do, but this is how it is done in kde5/ecm.eclass for years

dev-qt/qdbus/qdbus-5.15.0_beta4.ebuild
a17r: right, I was too fast there

eclass/qt5-build.eclass
Pesa: well, variables are cheap :) but the current way is perfectly fine with me too.

Chiitoo:
Looks good to me too so far, building-wise.  Did not look into things deeper than that yet though.

a17r:
> Github doesn't like your signature on these commits. Other than that, LGTM.

Thanks for the reminder, key updated.

> Looks good to me too so far, building-wise. Did not look into things deeper than that yet though.

overlay is for testing, right? ;)

PR #225, opened 2020-07-02
Phase out dev-qt/qtchooser

a17r:
At the same time we can discuss if/how to replace it.

EDIT 2021-01-10: Remaining list of revdeps: https://qa-reports.gentoo.org/output/genrdeps/rindex/dev-qt/qtchooser

### DIFF
--- a/sets/qt5-tools
+++ b/sets/qt5-tools
@@ -1,10 +1,10 @@
-dev-qt/assistant:5
+dev-qt/assistant:0
 dev-qt/designer:5
-dev-qt/linguist:5
+dev-qt/linguist:0
 dev-qt/linguist-tools:5
-dev-qt/pixeltool:5
-dev-qt/qdbus:5
-dev-qt/qdbusviewer:5
+dev-qt/pixeltool:0
+dev-qt/qdbus:0
+dev-qt/qdbusviewer:0
 dev-qt/qdoc:5
 dev-qt/qtdiag:5
 dev-qt/qthelp:5


Pesa:
hmm why? This should finally start to be useful with the Qt 5->6 transition...

a17r:
> hmm why? This should finally start to be useful with the Qt 5->6 transition...

We were wondering about that. Because upstream apparently considers it 'somewhat obsolete'. It is only providing links to standard system paths for the chosen Qt version, do any packages depend on that? At least for building Qt6 packages we won't need it, as we use eqmake{4,5,6} for that, and had fixed other build systems using qmake-utils.eclass in the past when they had been relying on qt binaries in PATH.

Pesa:
> We were wondering about that. Because upstream apparently considers it 'somewhat obsolete'.

Any references?

> It is only providing links to standard system paths for the chosen Qt version,

Not exactly. qtchooser is a multiplexing tool. e.g. you can run `qtpaths -qt=6` and get the Qt6 version of the executable. Or you can set the `QT_SELECT` env variable.

> do any packages depend on that? At least for building Qt6 packages we won't need it, as we use eqmake{4,5,6} for that, and had fixed other build systems using qmake-utils.eclass in the past when they had been relying on qt binaries in PATH.

Gentoo packages should NOT depend on qtchooser behavior, they should discover the appropriate Qt installation prefix through whatever means (this is implicit if using eqmake) and use the binaries from that prefix. Of course I'm sure there are buggy packages that don't do this correctly and will break when the default (`/etc/xdg/qtchooser/default.conf`) changes.

qtchooser and `QT_SELECT` are intended for users' consumption.

Chiitoo:
This started with a user (amadio) asking about 'qtconfig' not working, which lead to me noticing that it was removed from 'qttools' back in 2014, and that 'qtchooser' was last updated back in 2018.

At that time I didn't really have a grasp of what and how 'qtchooser' is to be used, but got some pointers from qt-labs IRC channel, so I guess it indeed can still be useful to users.

Today I finally got to asking the author of 'qtchooser' about this, and they explained that the old tools can not be removed due to older versions still having them, and they may be in use.

As for Qt 6, they do not plan to update 'qtchooser' for it, and instead, "need to adopt non-conflicting names or paths".

So with all this in mind, should our Qt packages still pull 'qtchooser' in or not?

a17r:
> Not exactly. qtchooser is a multiplexing tool. e.g. you can run `qtpaths -qt=6` and get the Qt6 version of the executable. Or you can set the `QT_SELECT` env variable.

I was actually aware of that and then conveniently shortened its abilities, you're right.

> Gentoo packages should NOT depend on qtchooser behavior, they should discover the appropriate Qt installation prefix through whatever means (this is implicit if using eqmake) and use the binaries from that prefix. Of course I'm sure there are buggy packages that don't do this correctly and will break when the default (`/etc/xdg/qtchooser/default.conf`) changes.
> 
> qtchooser and `QT_SELECT` are intended for users' consumption.

The validity of this RDEPEND is the main focus here then: https://github.com/gentoo/qt/blob/5c76498b7f0c2d4bdccc9110d53655accac74a38/eclass/qt5-build.eclass#L116

Chiitoo:
This just in [1].

> Thiago Macieira:
> As the qtchooser maintainer. I am declaring it EOL and have closed the 
> remaining open tasks.
> 
> I hereby also resign maintaining it too.

1. https://lists.qt-project.org/pipermail/development/2020-November/040589.html

Pesa:
Yeah I saw that whole thread. Upstream seems to finally understand that version-suffixed binaries are a better approach, although I'm not sure if the maintainers actually reached consensus on doing the renaming upstream (it's quite late in the development cycle) or they're letting each distro take care of it (which would suck).

Either way, qtchooser is definitely dead.

The question is what to do with existing Qt5 installations, which we still need to support for the foreseeable future. A few options:
1. Do nothing. The unversioned symlinks in `/usr/bin` are still owned by qtchooser and will always redirect to the Qt5 binaries in `/usr/lib/qt5/bin` (`QT5_BINDIR`).
2. Migrate away from qtchooser, *without* changing `QT5_BINDIR`. The symlinks in `/usr/bin` are installed by the qt package that installs the respective tool in `QT5_BINDIR`, e.g., qtcore installs `/usr/bin/qmake`. Symlinks can be (two sub-cases):
  2a. unversioned, e.g., `/usr/bin/qmake` -> `/usr/lib/qt5/bin/qmake`
  2b. versioned, e.g., `/usr/bin/qmake5` (or `qmake-qt5`) -> `/usr/lib/qt5/bin/qmake`
3. Migrate away from qtchooser *and* change `QT5_BINDIR` to `/usr/bin`. Again two sub-cases:
  3a. tools are installed as is, unversioned
  3b. tools are renamed to append either `-qt5` or just `5`
4. Something else?

a17r:
Pushed a solution following `2b.`: Keeps compatibility with `qmake-utils.eclass`, does not conflict with `qtchooser` paths should anyone still rely on it (would we move installing qtchooser configuration file into its ebuild or forget about it right away?). I tend to prefer `5` suffixing rather than oddly duplicating `-qt5`.

a17r:
Can't push Qt 5.15.2 right now anyway because once again, QtWebEngine does not build - neither 5.15.2 nor 5.15.9999, with different errors....

a17r:
> 4.: Symlink all binaries or check with upstream's recommendations [1]

Going forward here, assembling the list of binaries to symlink-with-suffix according to the upstream discussion linked above:
- qmake
- qml
- qtdiag
- qtpaths
- assistant
- designer

Finally, binaries we could(?) install unversioned in `/usr/bin` directly:
 - linguist
 - qdbus
 - qdbusviewer

Out of them only `dev-qt/qdbus` has few revdeps which should probably be converted to depend on it via `:*`?

Also mentioned in the discussion, but ultimately declared (Qt-)developer-only so not necessary to be in PATH:
- pixeltool
- qtplugininfo
- qtwaylandscanner
- tracegen
- sdpscanner

@pesa @kensington @Chiitoo what are your thoughts on this?

Qt 5.15.2 (minus jumbo-build...) could be moved to Gentoo ebuild repo very quickly once we have decided this.

Chiitoo:
Something going wrong-like with jumbo-build?

a17r:
Yup: https://bugreports.qt.io/browse/QTBUG-88657

Chiitoo:
I see.  Have not bumped into that myself.

As for all the rest, I'll trust the judgement of you and the others.  I can't think of anything wrong or anything to add.

a17r:
So, implemented all of https://github.com/gentoo/qt/pull/225#issuecomment-731724790 with symlinks now because that was the cheapest way to do it.

a17r:
Some numbers: Rebuilt 102 non-KDE revdeps of dev-qt/qtcore on my system.
- Missing moc in PATH: https://bugs.gentoo.org/show_bug.cgi?id=756262
- Missing lrelease in PATH: Easily fixed in b0d93f19dd63cf0dd6f9154ef33ef8326e3c7872 and 5d8db0c6c8a1cc7556c66b3ade0bbf959ebe71b8

Probably not too bad but for sure there will be some additional bugs like that.

a17r:
@Pesa do you want more links? should link creation be rather done in the respective ebuilds?

a17r:
Also:
- Missing xmlpatterns in PATH: https://bugs.gentoo.org/755977

Pesa:
> Pushed a solution following `2b.`

That was one of my least favorite, as it will (likely) break things for users and other packages, and I'm not sure if it's worth it now that Qt5 is essentially in maintenance mode... But I won't stop you if you want to deal with the fallout ;)
Moreover, while I also prefer just "5" as suffix, it won't help much at this point because: 1) it's a breaking change within gentoo as I said above, 2) it's not consistent with what other major distros have been using ("-qt5"), 3) it's also different from vanilla binaries from upstream which are not versioned. The only advantage is that it would be consistent with Qt6, but that's a pretty weak motivation IMHO.
That being said, it's ultimately up to you since you're doing all the work :)

> @Pesa should link creation be rather done in the respective ebuilds?

I'd prefer that, yes.

Pesa:
> Going forward here, assembling the list of binaries to symlink-with-suffix according to the upstream discussion linked above:
> 
> * qmake
> * qml
> * qtdiag
> * qtpaths
> * designer

Agree with the above.

> Considering it is going to be no [longer part of Qt6](https://lists.qt-project.org/pipermail/development/2020-November/040670.html), maybe install it un-suffixed:
> 
> * assistant
> 
> Finally, binaries we could(?) install w/o suffix in `/usr/bin` directly:
> 
> * linguist
> * qdbus
> * qdbusviewer
> 
> Out of them only `dev-qt/qdbus` has few revdeps which should probably be converted to depend on it via `:*`?

Or unSLOT those packages. Users should just use the latest and greatest and not care if it's Qt5- or Qt6- or Qt7-based. And there should not be any reasons to keep multiple versions simultaneously installed. I'm fine either way though.

> Also mentioned in the discussion, but ultimately declared (Qt-)developer-only so not necessary to be in PATH:
> 
> * pixeltool
> * qtplugininfo
> * qtwaylandscanner
> * tracegen
> * sdpscanner

It seems to me the ML discussion has been somewhat inconsistent wrt the criterion to apply to decide whether a binary should be in PATH or not. Sometimes they use "developer tool vs user-facing tool", other times "typically/often run manually by a human vs typically run by a build system or by some other program". IMHO the latter criterion is the correct one, but ultimately let's be consistent with other distros and potentially add a few more symlinks if they make our life easier.

> Missing xmlpatterns in PATH: https://bugs.gentoo.org/755977

Applying the same criterion used for assistant, I'd suggest to leave it alone (i.e., unversioned symlink in PATH), since there won't be a Qt6 version AFAIK.

a17r:
> > Pushed a solution following `2b.`
> 
> That was one of my least favorite, as it will (likely) break things for users and other packages, and I'm not sure if it's worth it now that Qt5 is essentially in maintenance mode... But I won't stop you if you want to deal with the fallout ;)

a) `3*` would have caused the most churn overall
b) I disliked `2|3a.` as it may collide with older Qt (in case people are still crazy enough to build/run that)

> Moreover, while I also prefer just "5" as suffix, it won't help much at this point because: 1) it's a breaking change within gentoo as I said above, 2) it's not consistent with what other major distros have been using ("-qt5"), 3) it's also different from vanilla binaries from upstream which are not versioned. The only advantage is that it would be consistent with Qt6, but that's a pretty weak motivation IMHO.

c) Any suffix in PATH will break assumptions - but that can be alleviated by still providing qtchooser for manual install and keeps b) alive; will need news item.

Yup, I like consistency, but we can still easily change the suffix until the last minute and we are in no hurry after I pushed 5.15.2 already to tree anyway. Can't hurt in case upstream still change their mind or flesh out more details.

> > Out of them only `dev-qt/qdbus` has few revdeps which should probably be converted to depend on it via `:*`?
> 
> Or unSLOT those packages. Users should just use the latest and greatest and not care if it's Qt5- or Qt6- or Qt7-based. And there should not be any reasons to keep multiple versions simultaneously installed. I'm fine either way though.
> 
> > Missing xmlpatterns in PATH: https://bugs.gentoo.org/755977
> 
> Applying the same logic used for assistant, I'd suggest to leave it alone (i.e., unversioned symlink in PATH), since there won't be a Qt6 version AFAIK.

All 3 `qdbus` revdeps are maintained by KDE proj so I can easily take care of that.
If we'd also unSLOT `qtxmlpatterns` - for consistency - that will be slightly messier with 59 ebuilds containing that string.

> It seems to me the ML discussion has been somewhat inconsistent wrt the criterion used to decide whether a binary should be in PATH or not. Sometimes they use "developer tool vs user-facing tool", other times "typically/often run manually by a human vs typically run by a build system or by some other program". IMHO the latter criterion is the correct one, but ultimately let's be consistent with other distros and potentially add a few more symlinks if they make our life easier.

Indeed, they should have had defined those terms beforehand.

a17r:
https://bugreports.qt.io/browse/QTBUG-89170

a17r:
I should add that toralf's tinderbox has been running without `dev-qt/qtchooser` (he put it in package.provided) for the past several days, which has only brought up a handful of fringe packages added to the qtbindir tracker. Two of those actually even supported the `5` suffix made in this PR (so I couldn't initially reproduce the failure), all of them were easily fixed.

a17r:
Update: https://lists.qt-project.org/pipermail/development/2021-February/040998.html
Moving build tools to libexec: https://bugreports.qt.io/browse/QTBUG-88791

eclass/qt5-build.eclass
```diff
@@ -101,7 +101,16 @@ fi
 : ${QT5_BUILD_DIR:=${S}_build}
 
 LICENSE="|| ( GPL-2 GPL-3 LGPL-3 ) FDL-1.3"
-SLOT=5/$(ver_cut 1-2)
+
+case ${PN} in
+	assistant|linguist|qdbus|qdbusviewer)
+		SLOT=0 ;;
+	linguist-tools|qtdiag|qtimageformats|qtpaths|qttranslations)
```
Pesa, line 108: qtgraphicaleffects and qtquickcontrols should be here too (pure QML modules)

and what about qtplugininfo?

eclass/qt5-build.eclass
```diff
@@ -101,7 +101,16 @@ fi
 : ${QT5_BUILD_DIR:=${S}_build}
 
 LICENSE="|| ( GPL-2 GPL-3 LGPL-3 ) FDL-1.3"
-SLOT=5/$(ver_cut 1-2)
+
+case ${PN} in
+	assistant|linguist|qdbus|qdbusviewer)
```
Pesa, line 106: pixeltool should be here, if not dead in qt6. If dead, it should be in the next category (SLOT=5).

Pesa (commented):
This looks great! How much stuff is failing without `qmake` in PATH?

eclass/qt5-build.eclass
Pesa: actually, I think it should be SLOT=0 either way (and symlinked since in Qt6 it's treated as a user-facing app)

eclass/qt5-build.eclass
Pesa: qtplugininfo is treated as user-facing in Qt6, but I don't know if Qt6 plugininfo can read Qt5 metadata (and vice versa).

dev-qt/qtchooser/files/qtchooser-66-qt-5.15.3-compat.patch
```diff
@@ -0,0 +1,33 @@
...
+ 	qdoc \
+ 	qdoc3 \
+@@ -44,9 +40,7 @@
+ 	rcc \
+ 	repc \
+ 	uic \
+-	uic3 \
+-	xmlpatterns \
+-	xmlpatternsvalidator \
++	uic3
```
Pesa, line 30: would it make sense to also drop the obsolete symlinks while at it, such as uic3, qml1plugindump, etc.? I can get a list if you think it's worthwhile

a17r: iirc I made some research for those binaries as for when they got obsolete, and I didn't want to drop them for at least Qt4 users (maybe even Qt3 but I have no idea if qtchooser was even relevant for them, and one could argue someone interested in that should use TQt3 instead these days).

a17r:
> This looks great! How much stuff is failing without `qmake` in PATH?

afaik toralf has been (and still is, I guess) running his tinderbox with qtchooser in package.provided since January 2021, without additional fallout. I see two cases that we would not catch with build errors:

a) ebuilds with `IUSE={qt5,gui}` suddenly ineffective silently
b) ebuilds assuming they build a qt5 GUI unconditionally by providing the dependency, but buildsys automagic drops it silently

These two possibilities we need to make maintainers aware of.

dev-qt/qtchooser/files/qtchooser-66-qt-5.15.3-compat.patch
Pesa: yeah, makes sense. I was concerned about collisions with Qt6 binaries in `/usr/bin` (such as `qmljs`) but I guess those will have a "6" suffix, so ignore me.

eclass/qt5-build.eclass
a17r: > qtgraphicaleffects and qtquickcontrols should be here too (pure QML modules)

Both are installing so files although in some private subdirectory... no headers though, so I guess this is fine.

a17r: > qtplugininfo is treated as user-facing in Qt6, but I don't know if the Qt6 tool can read Qt5 metadata (and vice versa).

Let's keep it at slot 5 then while we are unsure.

eclass/qt5-build.eclass
```diff
@@ -316,6 +326,16 @@ qt5-build_pkg_postrm() {
...
 ######  Public helpers  ######
 
+# @FUNCTION: qt5_symlink_binary_to_path
+# @USAGE: <target binary name> [suffix]
+# @DESCRIPTION:
+# Symlink a given binary from QT5_BINDIR to QT5_PREFIX/bin, with optional suffix
+qt5_symlink_binary_to_path() {
+	[[ $# -ge 1 ]] || die "${FUNCNAME}() requires at least one argument"
+
+	dosym -r "${QT5_BINDIR#${EPREFIX}}"/${1} /usr/bin/${1}${2}
```
thesamesam, line 336: Do we actually need the #$...? It figures out the shortest path anyway? Or what am I missing?

thesamesam (approved):

a17r:
5.15.3 version bump based on this PR is now almost ready: https://github.com/a17r/qt/commit/1b6163dfbc9f8090b43a1ea7f75e81a7e0a1c98d

a17r:
gentoo-dev mail draft: https://gist.github.com/a17r/b622b0be3e1e613f1f45c0b35c5196ee

eclass/qt5-build.eclass
```diff
@@ -125,14 +125,21 @@ fi
 
 LICENSE="|| ( GPL-2 GPL-3 LGPL-3 ) FDL-1.3"
 
-case ${PN} in
-	assistant|linguist|qdbus|qdbusviewer|pixeltool)
-		SLOT=0 ;;
-	linguist-tools|qtdiag|qtgraphicaleffects|qtimageformats|qtpaths| \
-	qtplugininfo|qtquickcontrols|qttranslations|qtwaylandscanner)
-		SLOT=5 ;;
+case ${PV} in
```
Pesa, line 128: a small formality: can you squash this commit with the earlier commit titled "qt5-build.eclass: Change SLOTing" since it's just a fixup?

dev-qt/qtdeclarative/qtdeclarative-5.15.3.9999.ebuild
```diff
@@ -56,3 +56,8 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+	qt5_symlink_binary_to_path qml 5
```
Pesa, line 62: I believe there's a few more here... in Qt6 the following are considered user-facing:

```
$ git grep qt_internal_add_app
tools/qml/CMakeLists.txt:qt_internal_add_app(qml
tools/qmleasing/CMakeLists.txt:qt_internal_add_app(qmleasing
tools/qmljs/CMakeLists.txt:qt_internal_add_app(qmljs
tools/qmlpreview/CMakeLists.txt:qt_internal_add_app(qmlpreview
tools/qmlscene/CMakeLists.txt:qt_internal_add_app(qmlscene
```

(`qmlscene` is deprecated in favor of `qml` in Qt6, but that's none of our business)

I don't know if the same tools should also be considered user-facing (and thus symlinked with suffix) in Qt5, but it seems a good starting point?

(and no, the classification criteria still don't make any sense to me... why is qmljs user-facing? sure, it can be manually invoked by a developer using Qt, but the same applies to qmllint, qmlprofiler, and several others...)

eclass/qt5-build.eclass
Pesa: > > qtgraphicaleffects and qtquickcontrols should be here too (pure QML modules)
> 
> Both are installing so files although in some private subdirectory... no headers though, so I guess this is fine.

yes the `.so` files just provide the (compiled C++) implementation of the QML components. By "pure QML" I meant that the modules have a QML-only API, they don't have a C++ API.

Pesa: > > qtplugininfo is treated as user-facing in Qt6, but I don't know if the Qt6 tool can read Qt5 metadata (and vice versa).
> 
> Let's keep it at slot 5 then while we are unsure.

Agreed. Should we symlink it though?

eclass/qt5-build.eclass
a17r: sure, that was the plan ;)

dev-qt/qtdeclarative/qtdeclarative-5.15.3.9999.ebuild
a17r: I need to work on the draft as well to not confuse what we mean with user-facing... Anyway, yes let's symlink those others too then.

a17r:
Changes: Added '5' suffix to dev-qt/qtxmlpatterns as it should have been. Dropped that part of the compatibility patch from qtchooser.

a17r:
Pending Qt 5.15.3 version bump based on this: https://github.com/gentoo/gentoo/pull/24628

dev-qt/qtdeclarative/qtdeclarative-5.15.3.9999.ebuild
Pesa: well, upstream uses USER_FACING in their cmake files... but yes, I agree on aligning with whatever upstream does.

Pesa:
> Changes: Added '5' suffix to dev-qt/qtxmlpatterns as it should have been. Dropped that part of the compatibility patch from qtchooser.

Why? I thought the unsuffixed name made for sense for xmlpatterns binaries, it's a very similar case to qdbus.

a17r:
> > Changes: Added '5' suffix to dev-qt/qtxmlpatterns as it should have been. Dropped that part of the compatibility patch from qtchooser.
> 
> Why? I thought the unsuffixed name made for sense for xmlpatterns binaries, it's very similar to the `qdbus` case.

Because it is also installing headers, so we can't give it slot '0', which would make it somewhat an outlier if installed suffixless. But it wouldn't conflict with older Qt either, as I at first thought.

Pesa:
> Because it is also installing headers, so we can't give it slot '0', which would make it somewhat an outlier if installed suffixless.

I don't think that's a problem. `xmlpatterns` should be one of those tools that you just use the latest and greatest and not care which major version of Qt they're based on. I don't think there will ever be a Qt6 version of qtxmlpatterns since the entire module was deprecated years ago, but if it does happen, we can just stop building/installing the binaries of the Qt5 version and replace them with the Qt6 ones.

a17r:
Okay then. I will
1) drop suffix again from qtxmlpatterns
2) switch it to clean slot 5 since we do not expect any relevant changes

We may have had the same conversation in the past about this and I just forgot.

a17r:
...saying that, I check qtxmlpatterns dev branch and notice 2 changes:
https://invent.kde.org/qt/qt/qtxmlpatterns/-/commit/d7e64c698eea21073645cffc993c14b729366e59
https://invent.kde.org/qt/qt/qtxmlpatterns/-/commit/99a1ba699b84f8a3d810f7b2d4d8929ff28c7ec5

Pesa:
> ...saying that, I check qtxmlpatterns dev branch and notice 2 changes: https://invent.kde.org/qt/qt/qtxmlpatterns/-/commit/d7e64c698eea21073645cffc993c14b729366e59 https://invent.kde.org/qt/qt/qtxmlpatterns/-/commit/99a1ba699b84f8a3d810f7b2d4d8929ff28c7ec5

Yeah I'm not sure why they even bothered... the first change looks automated/scripted, the second one is a no-op.

eclass/qt5-build.eclass
```diff
@@ -124,7 +124,24 @@ fi
...
+	5.15.2*)
+		SLOT=5/$(ver_cut 1-2)
+		;;
+	*)
+		case ${PN} in
+			assistant|linguist|qdbus|qdbusviewer|pixeltool)
+				SLOT=0 ;;
+			linguist-tools|qtdiag|qtgraphicaleffects|qtimageformats|qtpaths| \
+			qtplugininfo|qtquickcontrols|qttranslations|qtwaylandscanner)
+				SLOT=5 ;;
```
Pesa, line 138: just noticed qdoc should probably be here too

Pesa: and qtquicktimeline (QML-only module, no headers)

PR #226, opened 2020-07-25
dev-qt/qtdeclarative: Trigger rebuild on dev-qt/qtgui subslot update

a17r:
Closes: https://bugs.gentoo.org/733674

5.14.2 and 5.15.0 in tree should receive revbumps for the same change, I guess.

### DIFF
--- a/dev-qt/qtdeclarative/qtdeclarative-5.15.9999.ebuild
+++ b/dev-qt/qtdeclarative/qtdeclarative-5.15.9999.ebuild
@@ -17,7 +17,7 @@ BDEPEND="${PYTHON_DEPS}"
 # qtgui[gles2-only=] is needed because of bug 504322
 DEPEND="
 	~dev-qt/qtcore-${PV}
-	~dev-qt/qtgui-${PV}[gles2-only=,vulkan=]
+	~dev-qt/qtgui-${PV}:5=[gles2-only=,vulkan=]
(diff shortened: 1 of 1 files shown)


Pesa:
Will this guarantee that portage schedules the qtdeclarative upgrade/rebuild before other packages?

a17r:
It is not entirely clear from the description in the bug if qtdeclarative had been 'left behind' or if kolourpaint and kate were scheduled for rebuild simply ahead of them. It can't be guaranteed that it will be scheduled ahead of other packages having a slot operator on dev-qt/qtgui, but it would have definitely been scheduled ahead of kolourpaint and kate which do not have that. It would also solve the 'left behind' case as portage would insist on the qtdeclarative rebuild-turned-upgrade.

a17r:
However, situations like these are apparently still possible (just got this from a user in #gentoo-kde):
Qt packages: https://dpaste.com/37UU9GRWW
emerge -uvaDN world: https://gist.github.com/DogHackit/14b20b859c2c65fa771066730423e02d

kde-frameworks/kwidgetsaddons was the package that failed in the first place, this is after the first update attempt. And most or all these kde-frameworks/* packages are bound to fail because qtbase modules are inconsistent. I believe only the previously proposed wall of blockers would have solved this; on the other hand Qt 5.15 is the last major update.

PR #227, opened 2020-08-07
dev-qt/qtbluetooth: Trigger rebuild on dev-qt/qtcore subslot update

a17r:
Uses private QtCore headers.

Package-Manager: Portage-3.0.1, Repoman-2.3.23


a17r:
I suspected that, yes. qtbase we fixed recently, which was the most important. This PR is a byproduct of https://bugs.gentoo.org/736150 even though the missing upgrade may have also uncovered a missed bump of minimum Qt version.

PR #228, opened 2020-11-17
dev-qt/qtnetwork: Fix macro redefinition of QT_LINKED_OPENSSL warning

a17r:
Closes: https://bugs.gentoo.org/652650

### DIFF
--- a/dev-qt/qtnetwork/qtnetwork-5.15.9999.ebuild
+++ b/dev-qt/qtnetwork/qtnetwork-5.15.9999.ebuild
@@ -66,3 +66,12 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+	# workaround for bug 652650
+	if use ssl; then
+		sed -e "/^#define QT_LINKED_OPENSSL/s/$/ true/" \
+			-i "${D}${QT5_HEADERDIR}"/Gentoo/${PN}-qconfig.h || die
+	fi
+}


dev-qt/qtnetwork/qtnetwork-5.15.9999.ebuild
```diff
@@ -66,3 +66,10 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+	# workaround for bug 652650
+	sed -e "/^#define QT_LINKED_OPENSSL/s/$/ $(usex ssl true false)/" \
```
Pesa, line 73: uhm, IIUC this issue, the macro should not be defined at all when `ssl` is disabled.

Pesa: what's in `/usr/include/qt5/QtNetwork/qtnetwork-config.h` if you build with `USE=-ssl`? (not on a gentoo machine right now)

Pesa: essentially what I'm suggesting is

```bash
if use ssl; then
    sed ... '/^#define QT_LINKED_OPENSSL/s/$/ true/' ...
fi
```

a17r: ...right. not much thinking went into this last night. Line is `#define QT_NO_LINKED_OPENSSL` w `USE=-ssl.`

Pesa (approved):
thanks

PR #229, opened 2020-11-22
dev-qt: Drop IUSE=webkit

a17r:
[qtwebkit-removal](https://bugs.gentoo.org/show_bug.cgi?id=qtwebkit-removal) is progressing well and it is probably time to remove any ties from `dev-qt` category.


Pesa (approved):
Agreed.

PR #230, opened 2020-11-28
dev-qt/qtnetwork: update patch for LibreSSL

steils:
Package-Manager: Portage-3.0.10, Repoman-3.0.2
Signed-off-by: Stefan Strogin <steils@gentoo.org>


a17r (approved):
Thanks, it applies, therefore I merge.

steils:
@a17r Thank you for the quick reply. Will you sync the patch in the main tree as well?

a17r:
can do.

PR #232, opened 2020-12-26
qt5-build.eclass cleanup

a17r:
- sync with gentoo.git dropping `dev-qt/qtchooser` from default `RDEPEND` but stop short of picking https://github.com/gentoo/gentoo/commit/072da15f03fea298ef12bfec3ea8ac95f85a15d5 (not necessary imo and so qt overlay users may test w/o Qt binaries in `PATH`)
- plus move qtchooser config from eclass to qtchooser ebuild, both picked from https://github.com/gentoo/gentoo/pull/18810


Pesa:
does it still make sense to keep a qtchooser-9999 ebuild?

a17r:
Good question. I would just keep it at 9999 for the time being until we've settled for which symlinks we have to drop.

Pesa (approved):
Fine with me

PR #234, opened 2021-02-15
dev-qt/qtsql: Add IUSE=mariadb

a17r:


dev-qt/qtsql/qtsql-5.15.2-r1.ebuild
```diff
@@ -41,17 +43,28 @@ QT5_GENTOO_PRIVATE_CONFIG=(
 
 PATCHES=( "${FILESDIR}/${P}-mysql-use-pkgconfig.patch" )
 
+src_prepare() {
+	qt5-build_src_prepare
+	if use mariadb; then
+		sed -e "/type.*pkgConfig.*mysqlclient/s/mysqlclient/libmariadb/" \
+		-i src/plugins/sqldrivers/configure.json || die
```
Pesa, line 50: can you indent this line since it's a continuation?

Pesa (commented):
the revbump should be in the second commit, no?

a17r:
> the revbump should be in the second commit, no?

Yes, would have gone to gentoo.git instead of course, but I hope with a recent fix to akonadi we can avoid pt.2 of this PR altogether. Supporting mariadb-connector-c here is slightly messy as there are several revdeps depending on qtsql[mysql] that would need change/figuring out.

Do you think it is worth keeping the pkgconfig patch? It is certainly much cleaner than running mysql_config.

Pesa:
> Do you think it is worth keeping the pkgconfig patch? It is certainly much cleaner than running mysql_config.

It would definitely be cleaner. OTOH if it's not broken, why bother? Up to you, I have no preference.

PR #237, opened 2021-04-14
dev-qt/qtwebengine: Move 'net-libs/nodejs' to BDEPEND

holgersson32644:
This fixes commit 2d6ef7e74ec98657341817ab7bd2663ad2679b2e, where nodejs
was added as a build-time dependency, but put into DEPEND.
See: https://devmanual.gentoo.org/general-concepts/dependencies/index.html
Package-Manager: Portage-3.0.18, Repoman-3.0.3
Signed-off-by: Nils Freydank <holgersson@posteo.de>

### DIFF
--- a/dev-qt/qtwebengine/qtwebengine-5.15.9999.ebuild
+++ b/dev-qt/qtwebengine/qtwebengine-5.15.9999.ebuild
@@ -84,13 +84,12 @@ RDEPEND="
 		~dev-qt/qtwidgets-${QTVER}
 	)
 "
-DEPEND="${RDEPEND}
+BDEPEND="${RDEPEND}
 	${PYTHON_DEPS}
-	>=app-arch/gzip-1.7
 	dev-util/gperf
 	dev-util/ninja
 	dev-util/re2c
-	net-libs/nodejs
+	net-libs/nodejs[inspector]
 	sys-devel/bison
 "
 


Pesa:
Thanks! I see several more DEPENDs that should probably be in BDEPEND instead (bison, ninja, ...), why special-casing nodejs?

a17r:
That's likely all of `DEPEND`, right? @Chiitoo any comments? ;)

Chiitoo:
I concur.

Perhaps we could add either 'ssl' or 'inspector' to 'nodejs' since 'ssl' seems to be required.  Chromium ebuild has 'inspector' but, I've not found out why.

holgersson32644:
> Thanks! I see several more DEPENDs that should probably be in BDEPEND instead (bison, ninja, ...), why special-casing nodejs?

I prefer to avoid nodejs and just didn't look at the other deps at all - fixed!

dev-qt/qtwebengine/qtwebengine-5.15.9999.ebuild
```diff
@@ -84,13 +84,12 @@ RDEPEND="
 		~dev-qt/qtwidgets-${QTVER}
 	)
 "
-DEPEND="${RDEPEND}
+BDEPEND="${RDEPEND}
```
a17r, line 87: that's not how `BDEPEND` works. Keep `DEPEND="${RDEPEND}`, and don't inherit either of them in `BDEPEND`.

a17r (changes requested):

a17r (commented):
Fixed while merging. :)

PR #238, opened 2021-04-17
Support kde.org.eclass snapshots and use it for live repository handling; fix eclassdoc

a17r:
Necessary `kde.org.eclass` changes were already pushed to Gentoo ebuild repository.

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -76,26 +76,17 @@ _QT5_P=${QT5_MODULE}-everywhere-src-${PV}
 
 inherit estack flag-o-matic toolchain-funcs virtualx
 
-HOMEPAGE="https://www.qt.io/"
-LICENSE="|| ( GPL-2 GPL-3 LGPL-3 ) FDL-1.3"
-SLOT=5/$(ver_cut 1-2)
-
-case ${PV} in
-	5.15.9999)
-		# KDE upstream for 5.15 patches
-		HOMEPAGE+=" https://invent.kde.org/qt/qt/"
-		EGIT_BRANCH="kde/5.15"
-		;;
-	*)
+if [[ ${PN} != qtwebengine ]]; then
+	if [[ ${QT5_BUILD_TYPE} == live ]] || [[ -n ${KDE_ORG_COMMIT} ]]; then
+		# KDE Qt5PatchCollection
+		inherit kde.org
+	else
 		# official stable release
+		HOMEPAGE="https://www.qt.io/"
 		SRC_URI="https://download.qt.io/official_releases/qt/${PV%.*}/${PV}/submodules/${_QT5_P}.tar.xz"
 		S=${WORKDIR}/${_QT5_P}
-		;;
-esac
-
-EGIT_REPO_URI=( "https://invent.kde.org/qt/qt/${QT5_MODULE}.git" )
-
-[[ ${QT5_BUILD_TYPE} == live ]] && inherit git-r3
+	fi
+fi
 
 # @ECLASS-VARIABLE: QT5_BUILD_DIR
 # @OUTPUT_VARIABLE
@@ -103,6 +94,8 @@ EGIT_REPO_URI=( "https://invent.kde.org/qt/qt/${QT5_MODULE}.git" )
 # Build directory for out-of-source builds.
 : ${QT5_BUILD_DIR:=${S}_build}
 
+LICENSE="|| ( GPL-2 GPL-3 LGPL-3 ) FDL-1.3"
+SLOT=5/$(ver_cut 1-2)
 IUSE="debug test"
 
 if [[ ${QT5_BUILD_TYPE} == release ]]; then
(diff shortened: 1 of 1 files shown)


Pesa:
Most commits in the series are ok (with minor comments I'll post later) but I'm not sure I follow the kde.org.eclass dependency... how much logic are we actually using from that eclass?

a17r:
In terms of the overlay, actually not much - the repository handling is trivial and src_unpack identical - but for using snapshots it will get interesting. So I mostly did it in order for the live ebuilds to stay compatible (copyable) with future "release" snapshots.

a17r:
> For using snapshots it will get interesting - it will just be copy from 5.15.9999 and set `KDE_ORG_COMMIT` value.

Speaking of which, do you want to make the snapshotty nature visible at all or are you fine with regular revbumps to the existing 5.15.2 versioning?

eclass/qt5-build.eclass
```diff
@@ -136,6 +136,17 @@ qt5-build_src_unpack() {
 qt5-build_src_prepare() {
 	qt5_prepare_env
 
+	if [[ -n ${KDE_ORG_COMMIT} ]]; then
+		# Upstream bumped version in 5.15 branch after 5.15.2 release but their
+		# 5.15.3 release is closed and this will never be more than a Qt 5.15.2
+		# with patches on top.
+		einfo "Preparing KDE Qt5PatchCollection snapshot at ${KDE_ORG_COMMIT}"
+		sed -e "/^MODULE_VERSION/s/5\.15\.3/5\.15\.2/" -i .qmake.conf || die
+		if ! [[ -d .git ]]; then
```
Pesa, line 150: no need for this conditional if you're using `mkdir -p` later, or am I missing something?

a17r:
This is now straightforward to test since kde.org.eclass changes were pushed to gentoo.git.

dev-qt/assistant/assistant-5.15.9999.ebuild
```diff
@@ -1,10 +1,10 @@
-# Copyright 1999-2020 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
 
 QT5_MODULE="qttools"
-inherit desktop qt5-build xdg-utils
+inherit desktop qt5-build xdg-utils kde.org
```
Pesa, line 7: I was wondering why `kde.org` is inherited by every single ebuild instead of by `qt5-build.eclass`?

eclass/qt5-build.eclass
```diff
@@ -74,16 +81,15 @@ case ${PV} in
 	*_alpha*|*_beta*|*_rc*)
 		# development release
 		QT5_BUILD_TYPE="release"
-		MY_P=${QT5_MODULE}-everywhere-src-${PV/_/-}
-		SRC_URI="https://download.qt.io/development_releases/qt/${PV%.*}/${PV/_/-}/submodules/${MY_P}.tar.xz"
-		S=${WORKDIR}/${MY_P}
+		_QT5_P=${QT5_MODULE}-everywhere-src-${PV/_/-}
```
Pesa, line 88: maybe you can use this definition for the default assignment above? since the stable versions won't have any `_` in `PV`, the substitution `${PV/_/-}` should be a noop.

or maybe just drop the whole "development release" case? I suppose there won't be any more

Pesa (changes requested):
The first two commits should be squashed IMHO as they really are one logical change.

Pesa:
> Speaking of which, do you want to make the snapshotty nature visible at all or are you fine with regular revbumps to the existing 5.15.2 versioning?

I'm not sure what you mean?

dev-qt/assistant/assistant-5.15.9999.ebuild
a17r: Can do, although every single ebuild it is not - e.g. dev-qt/qtwebengine is still pulled from qt.io, and not every module got changes pushed to invent.kde.org yet. It was just a way to make sure none of the existing ebuilds is affected by the eclass changes.

I suppose kde.org could be conditionally inherited if `KDE_ORG_COMMIT` is set.

Pesa:
> Repository/tarball handling is `kde.org.eclass`' single purpose. ;) 

Yeah I should've read the eclass before asking, sorry 😅  I'm ok with this approach.

dev-qt/assistant/assistant-5.15.9999.ebuild
Pesa: ok I see, I'm fine either way.

a17r:
> > Speaking of which, do you want to make the snapshotty nature visible at all or are you fine with regular revbumps to the existing 5.15.2 versioning?
> 
> I'm not sure what you mean?

Convention would make us create ebuilds with _pYYYYMMDD version postfix for snapshots.

dev-qt/assistant/assistant-5.15.9999.ebuild
a17r: Went a step further and made `qt5-build.eclass` fully embrace `kde.org.eclass` by dropping `src_unpack()` phase. I like that solution now much better.

Pesa:
> Convention would make us create ebuilds with `_pYYYYMMDD` version postfix for snapshots.

hmm good question. I think using `_pYYYYMMDD` would be slightly clearer, but if that's messy to support I can accept simple revbumps.

Btw, are we going to roll our own tarballs from git (à la `git archive`) or use the official 5.15.2 tarball + (a tarball of) patches?

a17r:
> hmm good question. I think using `_pYYYYMMDD` would be clearer, but if that's messy to support I can accept simple revbumps.

I'm not decided yet if `_pYYYYMMDD` is opening a can of worms. Up to now, all Qt ebuilds have been neatly aligned.

> Btw, are we going to roll our own tarballs from git (à la `git archive`) or use the official 5.15.2 tarball + (a tarball of) patches?

`KDE_ORG_COMMIT` is rolling our own. The problem with `5.15.2` tarball is that it was branched off `5.15` and diverged from it before release. That makes every odd patch of the series fail and is no fun when there are >170 already. I did not pursue it further but when looking at the Mandriva 5.15.3 package they left quite a few patches out and the first few I checked correlated with those failing against `5.15.2` tarball.

a17r:
> > hmm good question. I think using `_pYYYYMMDD` would be clearer, but if that's messy to support I can accept simple revbumps.
> 
> I'm not decided yet if `_pYYYYMMDD` is opening a can of worms. Up to now, all Qt ebuilds have been neatly aligned. It would complicate simple bugfix bumps of a single module.

To elaborate on that, we will need to change all Qt ebuilds' `dev-qt/` deps to something like
- `~dev-qt/qtsameqtmodule-${PV}`
- `=dev-qt/qtotherqtmodule-${ver_cut 1-3}*`

Which brings potential minor trouble for the odd revdep with weird `dev-qt/` dependency version defs and more headaches should it happen that we need or want to have e.g. KDE packages depend on something beyond 5.15.2 where a single >=version is not good enough anymore. Arguably the latter would be even worse if we only did revbumps though...

Pesa:
Right. It's ugly either way. For simplicity, I'd suggest starting with revision bumps(*) and see how the situation evolves.

(*) we could use a (potentially silly) numbering scheme where we bump the revision number to the next multiple of 10 when we update to a newer "upstream" snapshot, e.g.:
* 5.15.2-r10 first snapshot from the Qt5PatchCollection
* 5.15.2-r11, -r12, -r13, ... ebuild fixes or other hotfixes on top of the -r10 snapshot
* 5.15.2-r20 second snaphot from the Qt5PatchCollection
* and so on...

> The other question is if we should be playing it safe by bumping the subslot, even though we keep the PV at 5.15.2...

Let's see how it goes without bumping it and if there are widespread issues we can reconsider... what do you say?

PR #240, opened 2021-08-01
Update Readme

damobrisbane:
I found the README a little cryptic just to "get started". Took https://github.com/damobrisbane/jorgicio-gentoo-overlay/edit/main/README.md as an example of using layman to init repo. Their example of using layman seems ok.


Pesa:
Thanks

PR #241, opened 2021-08-19
qt5-build.eclass: Support EAPI-8

a17r:
I don't see anything else to change.

### DIFF
--- a/dev-qt/qtdeclarative/qtdeclarative-5.15.9999.ebuild
+++ b/dev-qt/qtdeclarative/qtdeclarative-5.15.9999.ebuild
@@ -1,9 +1,9 @@
 # Copyright 2009-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-PYTHON_COMPAT=( python3_{6..9} )
+PYTHON_COMPAT=( python3_{8..10} )
 inherit python-any-r1 qt5-build
 
 DESCRIPTION="The QML and Quick modules for the Qt5 framework"
@@ -14,7 +14,6 @@ fi
 
 IUSE="gles2-only +jit localstorage vulkan +widgets"
 
-BDEPEND="${PYTHON_DEPS}"
 # qtgui[gles2-only=] is needed because of bug 504322
 DEPEND="
 	~dev-qt/qtcore-${PV}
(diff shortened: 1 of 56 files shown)


Pesa (commented):
`RESTRICT+=...` is no longer needed in EAPI 8 because the variable is now accumulated automatically... but I suppose it doesn't hurt either?

a17r:
We could change that with dropping EAPI-7.

a17r:
> `RESTRICT+=...` is no longer needed in EAPI 8 because the variable is now accumulated automatically... but I suppose it doesn't hurt either?

Appended EAPI-8 bumps to dev-qt/* and dropped EAPI-7 support in subsequent commit.

dev-qt/qtdeclarative/qtdeclarative-5.15.9999.ebuild
```diff
@@ -1,9 +1,9 @@
 # Copyright 2009-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
-EAPI=7
+EAPI=8
 
-PYTHON_COMPAT=( python3_{6..9} )
+PYTHON_COMPAT=( python3_{7..10} )
```
Pesa, line 6: 3.7 is gone so you can start from 8

dev-qt/qtnetwork/qtnetwork-5.15.9999.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 1999-2021 Gentoo Authors
```
Pesa, line 1: (stupid github doesn't let me add comments on lines outside the immediate context of the diff...)

Around lines 55-60 there are 3 cases of `usex` that can be turned into `usev` dropping the third (empty) argument.

dev-qt/qtsql/qtsql-5.15.9999.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 1999-2021 Gentoo Authors
```
Pesa, line 1: another case of `usex` -> `usev` below

dev-qt/qtwidgets/qtwidgets-5.15.9999.ebuild
```diff
@@ -1,7 +1,7 @@
 # Copyright 1999-2021 Gentoo Authors
```
Pesa, line 1: another `usex` -> `usev` on line 54

Pesa (approved):
very minor comments, no need to amend these commits, you can push follow-up patches

and thanks a lot!

Pesa:
btw your signature shows up as "unverified" on github

PR #245, opened 2021-09-03
qt5-build.eclass: Set BDEPEND=dev-qt/qtxml for QT5_MODULE=qttools

a17r:


Pesa:
Please see https://bugs.gentoo.org/728278#c10 and let me know what you think.

PR #249, opened 2021-09-22
dev-qt/qtwidgets: Add IUSE=dbus to enable xdgdesktopportal build

a17r:
New use flag builds Qt5 plugin for integration with flatpak and snap.
    
Add missing IUSE=gtk dependencies. No additional dependencies with
IUSE=dbus means the added REQUIRED_USE is cheap, and sys-apps/dbus is
already pulled in by x11-libs/gtk+3 (via dev-libs/glib) unconditionally.

### DIFF
--- a/dev-qt/qtwidgets/qtwidgets-5.15.2.9999.ebuild
+++ b/dev-qt/qtwidgets/qtwidgets-5.15.2.9999.ebuild
@@ -13,13 +13,17 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
 fi
 
 # keep IUSE defaults in sync with qtgui
-IUSE="gles2-only gtk +png +X"
+IUSE="dbus gles2-only gtk +png +X"
+
+REQUIRED_USE="gtk? ( dbus )"
 
 DEPEND="
 	=dev-qt/qtcore-${QT5_PV}*:5=
-	=dev-qt/qtgui-${QT5_PV}*[gles2-only=,png=,X?]
+	=dev-qt/qtgui-${QT5_PV}*:5=[gles2-only=,png=,X?]
+	dbus? ( =dev-qt/qtdbus-${QT5_PV}* )
 	gtk? (
-		=dev-qt/qtgui-${QT5_PV}*[dbus]
+		dev-libs/glib:2
+		=dev-qt/qtgui-${QT5_PV}*:5=[dbus]
 		x11-libs/gtk+:3
 		x11-libs/libX11
 		x11-libs/pango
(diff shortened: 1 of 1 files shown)


dev-qt/qtwidgets/qtwidgets-5.15.2.9999.ebuild
```diff
@@ -13,12 +13,16 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
...
 
 # keep IUSE defaults in sync with qtgui
-IUSE="gles2-only gtk +png +X"
+IUSE="dbus gles2-only gtk +png +X"
+
+REQUIRED_USE="gtk? ( dbus )"
 
 DEPEND="
 	=dev-qt/qtcore-${QT5_PV}*:5=
 	=dev-qt/qtgui-${QT5_PV}*[gles2-only=,png=,X?]
```
Pesa, line 22: while at it, please add a subslot dependency here

dev-qt/qtwidgets/qtwidgets-5.15.2.9999.ebuild
```diff
@@ -13,12 +13,16 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
...
+	dbus? ( =dev-qt/qtdbus-${QT5_PV}*:5= )
```
Pesa, line 23: and remove it from here, "dbus-private" is not used in xdgdesktopportal

dev-qt/qtwidgets/qtwidgets-5.15.2.9999.ebuild
```diff
@@ -13,12 +13,16 @@ if [[ ${QT5_BUILD_TYPE} == release ]]; then
...
 	gtk? (
+		dev-libs/glib:2
 		=dev-qt/qtgui-${QT5_PV}*[dbus]
```
Pesa, line 26: and add it here :)

Pesa (changes requested):
I don't remember why the gtk3 plugin depends on dbus... can you enlighten me? might be worth adding a comment in the ebuild if the reason isn't straightforward...

a17r:
> I don't remember why the gtk3 plugin depends on dbus... can you enlighten me? might be worth adding a comment in the ebuild if the reason isn't straightforward...

I don't know either, qmake being much of a black box to me. I'm going by lddtree again:
```
libqgtk3.so => /usr/lib64/qt5/plugins/platformthemes/libqgtk3.so (interpreter => none)
    libQt5DBus.so.5 => /usr/lib64/libQt5DBus.so.5
        libpthread.so.0 => /lib64/libpthread.so.0
            ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
        libdbus-1.so.3 => /usr/lib64/libdbus-1.so.3
```

PR #250, opened 2022-02-17
dev-qt/qtwaylandscanner split

a17r:
Sort of a messy split, but I did not quite know how else to do it.

Ideally we would be able to split off the compositor but that is rather tricky.

### DIFF
--- a/dev-qt/qtwayland/qtwayland-5.15.3.9999.ebuild
+++ b/dev-qt/qtwayland/qtwayland-5.15.3.9999.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2022 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=8
@@ -40,4 +40,9 @@ src_configure() {
 		$(qt_use X feature-xcomposite-glx)
 	)
 	qt5-build_src_configure
+}
+
+src_install() {
+	qt5-build_src_install
+	rm "${D}${QT5_BINDIR}"/qtwaylandscanner || die
 }
--- a/dev-qt/qtwaylandscanner/qtwaylandscanner-5.15.3.9999.ebuild
+++ b/dev-qt/qtwaylandscanner/qtwaylandscanner-5.15.3.9999.ebuild
@@ -1,0 +1,24 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+QT5_MODULE="qtwayland"
+inherit qt5-build
+
+DESCRIPTION="Tool that generates certain boilerplate C++ code from Wayland protocol xml spec"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~riscv ~sparc ~x86"
+fi
+
+SLOT=5
+
+DEPEND="=dev-qt/qtcore-${QT5_PV}*:5="
+RDEPEND="${DEPEND}
+	!<dev-qt/qtwayland-5.15.3:5
+"
+
+QT5_TARGET_SUBDIRS=(
+	src/qtwaylandscanner
+)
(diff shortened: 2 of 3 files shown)


Pesa:
> Sort of a messy split, but I did not quite know how else to do it.

I've seen worse ;)
But what's the reason for splitting?

dev-qt/qtwaylandscanner/qtwaylandscanner-5.15.2.9999.ebuild
```diff
@@ -0,0 +1,22 @@
...
+DESCRIPTION="Wayland platform plugin for Qt"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~riscv ~sparc ~x86"
+fi
+
+SLOT=5
+
+DEPEND="=dev-qt/qtcore-${QT5_PV}*:5="
+RDEPEND="${DEPEND}"
```
Pesa, line 18: do we need to block older qtwayland, in case of downgrades?

dev-qt/qtwaylandscanner/qtwaylandscanner-5.15.2.9999.ebuild
```diff
@@ -0,0 +1,22 @@
+# Copyright 1999-2022 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+QT5_MODULE="qtwayland"
+inherit qt5-build
+
+DESCRIPTION="Wayland platform plugin for Qt"
```
Pesa, line 9: put something more appropriate here please

a17r:
> But what's the reason for splitting?

Several revdeps do not need the full qtwayland at all, and qtwaylandscanner would be in BDEPEND analogous to wayland-scanner. Afaik barely anything makes use of qtwaylandcompositor, which would be really nice to be able to split off.

dev-qt/qtwayland/qtwayland-5.15.3.9999.ebuild
```diff
@@ -27,7 +27,9 @@ DEPEND="
 		x11-libs/libXcomposite
 	)
 "
-RDEPEND="${DEPEND}"
+RDEPEND="${DEPEND}
+	=dev-qt/qtwaylandscanner-${QT5_PV}*:5
```
a17r, line 31: We could use the [5.15.3 bump](https://github.com/gentoo/qt/pull/251) as a way to communicate this dependency being dropped.

dev-qt/qtwayland/qtwayland-5.15.3.9999.ebuild
```diff
@@ -41,3 +43,8 @@ src_configure() {
 	)
 	qt5-build_src_configure
 }
+
+src_install() {
+	qt5-build_src_install
+	rm "${D}"/usr/$(get_libdir)/qt5/bin/qtwaylandscanner || die
```
Pesa, line 49: use `QT5_BINDIR`?

dev-qt/qtwaylandscanner/qtwaylandscanner-5.15.3.9999.ebuild
```diff
@@ -0,0 +1,24 @@
...
+QT5_MODULE="qtwayland"
+inherit qt5-build
+
+DESCRIPTION="Tool that generates certain boilerplate C++ code from Wayland protocol xml spec"
+
+if [[ ${QT5_BUILD_TYPE} == release ]]; then
+	KEYWORDS="~amd64 ~arm ~arm64 ~hppa ~ppc ~ppc64 ~riscv ~sparc ~x86"
+fi
+
+SLOT=5
```
Pesa, line 15: I suppose this will move to the eclass later? (related to #225)

a17r: Would make sense yes.

dev-qt/qtwayland/qtwayland-5.15.3.9999.ebuild
a17r: Dropped dependency and will make it part of the initial package.mask message and/or gentoo-dev mail.

PR #252, opened 2022-04-17
Add Qt6 ready dev-qt/qt-creator-9999

Peter-Levine:
- **dev-qt/qttranslations: SLOT=6 initial commit** 

- **dev-qt/qt-creator: bump 9999**
Switched from the qmake to cmake buildsystem.  Added support for the
Coco code coverage tool. The 'git' USE flag now also builds the gitlab
plugin. The 'wayland' flag was added to enforce built-with-use
dependency. 'imageviewer' flag is added for conditional dependence on
dev-qt/qtsvg. clangd functionality now depends on >=clang-14.*.

Closes: https://github.com/gentoo/qt/pull/252
Signed-off-by: Peter Levine <plevine457@gmail.com>


AdelKS:
Hello!

Thanks for this work ! It compiles fine and well. I unmasked the Qt 6 packages so it pulled them off as a dependency, I suppose that's on purpose.

Something that is more important to me, did you find a way to enable static analysis of C++ 20 code ? Including for example `string_view` through `#include <string_view>` makes the code analysis complain. Probably because it's limited to C++17. I also got this message
```
Clang Code Model: Error: The clangbackend executable "/usr/libexec/qtcreator/clangbackend" does not exist.
```
In Qt creator 6.0 it works with a simple `CONFIG += console c++2a` in the `.pro` file.

AdelKS:
Okay the `ClangCodeModel` doesn't work without trying to enable C++ 20, it gives errors with this

`test-cpp.pro`
```
TEMPLATE = app
CONFIG += console
CONFIG -= app_bundle
CONFIG -= qt

SOURCES += \
        main.cpp
```
`main.cpp`
```c++
#include <iostream>

using namespace std;
int main()
{
    cout << "Hello world!" << endl;
    return 0;
}
```

Peter-Levine:
@AdelKS It looks like I might have overlooked the necessity of certain tools by placing all of them behind a default-disabled `tools` USE flag.  clandbackend should be governed by the `clang` USE flag, among other issues.  Please rebuild with `tools` enabled and confirm whether you can still observe the issue.

AdelKS:
Hello, enabling the `tools` use flag did not solve the issue. However, putting this in ` ~/.config/clangd`
```                                                                                                                                
CompileFlags:
  # Treat code as C++, use C++20 standard, enable more warnings.
  Add: [-xc++, -std=c++20, -Wall, -I/usr/lib/clang/14.0.4/include]
```
especially the `-I/usr/lib/clang/14.0.4/include`, solves it. Maybe the build can fix this, because it used to work in Qt Creator 6.0 without user intervention.

Peter-Levine:
Can confirm in **Tools -> Options -> C++ -> Clangd** that **Path to executable** is `/usr/lib/llvm/14/bin/clangd` and **Insert header files on completion** is checked?

Peter-Levine:
I have run into this: [QTCREATORBUG-27623: clangd says it can't find stddef.h](https://bugreports.qt.io/browse/QTCREATORBUG-27623?gerritReviewStatus=All).

AdelKS:
> Can you confirm in Tools -> Options -> C++ -> Clangd that Path to executable is /usr/lib/llvm/14/bin/clangd and Insert header files on completion is checked?

Yeah, it does not fix it by doing that alone. It looks like I really need to do the workaround I shown above by setting a file in the `.config` folder.

Peter-Levine:
> Hello, enabling the `tools` use flag did not solve the issue. However, putting this in ` ~/.config/clangd`

Sorry.  I was conflating things.


I'm able to reproduce the issue.  For some reason, `clangd is not finding clang's internal C standard headers.  It's not designed to use GCC's.  In the next few days, I'll try to narrow down the cause.

Peter-Levine:
OK. It should be fixed. Please test when you get a chance.

AdelKS:
It works now, thanks for figuring it out !

A note: I cannot change the path for `clangd` in the settings to the one version 14, after restarting Qt Creator it changes back to version 13, I suppose because Qt Creator doesn't work with it ?

Also, please do not default enable (nearly) all plugins. If you keep the defaults from the previous ebuild in ::gentoo that would already be great.

Peter-Levine:
clangd on my end is set to version 14.  It may be that you changed/added/removed a clang version?  If so, rebuild qt-creator and report if that resolves the issue.  Also, if any of the .config files were altered by root, the changes might be ephemeral without user write perms.

Defaulted USE flags are reverted to those of qt-creator-6.

AdelKS:
>  If so, rebuild qt-creator and report if that resolves the issue.

Rebuilding partly solved the issue:  `clangd` is disabled at every startup of Qt Creator, but I can activate it and change the path to 14. But these changes are still not saved, so I need to do it at each restart :thinking:

Peter-Levine:
Added `qt-creator-8.0.0_beta1.ebuild`.

> Weirdly enough, my configuration in the `~/.config` was borked, so I deleted it and all good now.

I reproduced the issue.  There appears to be a problem with the way the clangd path setting interacts with `~/.config/QtProject/QtCreator.ini`.  After changing the clangd version and saving, attempts to change it back fail to alter the `ClangdPath=` entry in `~/.config/QtProject/QtCreator.ini`.   Directly editing `~/.config/QtProject/QtCreator.ini` or deleting it and letting qt-creator recreate it are both effective workarounds.

It seems fixed in qt-creator-8 but only clangd from >=clang-14 is supported.  Attempts to set it to clang-13's clangd emits `The clangd version is 13.0.1, but 14 or greater is required.` just below the `Path to executable:` config option.

Peter-Levine:
Actually, `ClangdPath=` is messed up in qt-creator-8.0.0_beta1, too.

Peter-Levine:
The problem with qt-creator-8.0.0_beta1 is https://github.com/qt-creator/qt-creator/commit/23fceb9ef36a70ca1520ef45baff07a7af46aecd#diff-ef1f834cd227f9e22fae1b5fc32863f5fabd634f255b4d8fab7e36f8249fc3c1R408.  It looks for the clang includes relative to clangd at `../lib/clang/${clang_version}/include`.  Because in `sys-devel/clang`, Gentoo sets `-DCLANG_RESOURCE_DIR="../../../../lib/clang/${clang_version}"`, and AFAICT is the only distro to do so, the includes are actually `../../../../lib/clang/${clang_version}/include`. 

But from the [clangd FAQ](https://clangd.llvm.org/troubleshooting#cant-find-compiler-built-in-headers-stddefh-etc):
> They must be found in ../lib/clang/\<version\>/include. If you’ve built clangd from source or installed it from a package, these should be in the right place. If you’ve moved the binary around, move it back!

I could continue editing this out but it may be accepted as a `sys-devel/clang` bug so I'll likely file a bug later and decide how to proceed.

Peter-Levine:
https://bugs.gentoo.org/853967

Peter-Levine:
[[QTCREATORBUG-27760\] - clangd can't find includes if clang is build with nonstandard CLANG_RESOURCE_DIR](https://bugreports.qt.io/browse/QTCREATORBUG-27760).

We'll see what comes of it.

AdelKS:
Thanks for this great job! Could you reproduce the Wayland issue ? (I only tested under Gnome although I don't see why it should be Gnome specific)

Peter-Levine:
Yeah. It might require a qt6 version of `dev-qt/qtwayland`.  I'll test when I get a chance.

Peter-Levine:
Ok.  There is a configure option for `qtwayland:6` which is currently not supported in the ebuild, `-DFEATURE_wayland_libhybris_egl_server_buffer=ON`.  After enabling that option in  `CMakeCache.txt` after `ebuild ... configure` and then merging, the Wayland issue is resolved.

Peter-Levine:
Actually, the Wayland issue appears to be resolved on my end with `dev-qt/qt-base[egl]`.

Peter-Levine:
Wayland issues should be fixed.

installgentoo:
You should add perfprofiler? ( dev-libs/elfutils )

perfprofiler plugin is useless without them, as it can't access kernel perf thing.

installgentoo:
You do want to make elfutils be dependant on perfprof use flag, not tools. Tools don't have anything to do with the tools we use inside creator, those are standalone things that qt team doesn't even package into their own builds.

Peter-Levine:
I was under the mistaken impression that not setting any configure options for executables would cause them to default to `OFF`.  It appears that they all default to `ON` with the exception of `BUILD_CPLUSPLUS_TOOLS`.  So I'll leave  `BUILD_CPLUSPLUS_TOOLS` tied to the `tools` flag and I'll add a flag for `perfparser`.  buildoutputparser is dependent on `qmake` and qml2puppet on `qml`, so I'll put them behind those flags.  The rest have no dependencies and should be removed from the ebuild and default to `ON`.

Peter-Levine:
Actually, the 6.0.0 ebuild makes perfparser dependent on the `perfprofiler` flag.  This seems a better choice, in retrospect, as it doesn't introduce another USE flag.

Miezhiko:
Live version has dropped Qt5 support.

Peter-Levine:
I kept QT5 in to possibly push to main Gentoo repo but I'd still need to edit out all the qt6 stuff to prevent it being flagged for bad deps and remove the qml stuff that requires QT6.  But I think I'll just edit out the QT5 stuff along with the qt6 USE flag, instead, and simplify it.

Miezhiko:
> I kept QT5 in to possibly push to main Gentoo repo but I'd still need to edit out all the qt6 stuff to prevent it being flagged for bad deps and remove the qml stuff that requires QT6. But I think I'll just edit out the QT5 stuff along with the qt6 USE flag, instead, and simplify it.

I mean few days ago qt creator (master) disabled qt 5 support an it's different calls for qt6, impossible to build with qt5

Peter-Levine:
Yeah.  I see it's in master branch. Still, it makes my decision to get rid of QT5 support all that easier.

Peter-Levine:
Then again, there are many who run plasma and don't want to lug around two versions of the QT framework while waiting for Plasma 6, so I might as well keep QT5 support in 8.x for the time being and take it out for live ebuild.

Peter-Levine:
Closing in favor of https://github.com/gentoo/gentoo/pull/26675.

Miezhiko:
@Peter-Levine I personally think it will be great to have `9999` (live) version in this overlay, PR on gentoo is missing work with Qt6

AdelKS:
Hello,

Thanks for keeping this PR up-to-date, maybe @Chiitoo can merge it ?

I tried to emerge qt-creator-10.0.1 on my machine and it failed, here's the build log:

[qt-creator-10-build.log.txt](https://github.com/gentoo/qt/files/12504361/qt-creator-10-build.log.txt)

It looks that this is the relevant error message:
```
var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp: In member function ‘QVector4D QmlDesigner::Internal::GeneralHelper::focusNodesToCamera(QQuick3DCamera*, float, const QVariant&, QQuick3DViewport*, float, bool, bool)’:
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:218:21: error: ‘QSSGRef’ was not declared in this scope
  218 |                     QSSGRef<QSSGRenderContextInterface> context;
      |                     ^~~~~~~
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:218:55: error: expected primary-expression before ‘>’ token
  218 |                     QSSGRef<QSSGRenderContextInterface> context;
      |                                                       ^
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:218:57: error: ‘context’ was not declared in this scope; did you mean ‘connect’?
  218 |                     QSSGRef<QSSGRenderContextInterface> context;
      |                                                         ^~~~~~~
      |                                                         connect
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:222:57: error: ‘class QQuick3DSceneManager’ has no member named ‘rci’
  222 |                     context = targetPriv->sceneManager->rci;
      |                                                         ^~~
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp: In member function ‘bool QmlDesigner::Internal::GeneralHelper::getBounds(QQuick3DViewport*, QQuick3DNode*, QVector3D&, QVector3D&)’:
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:892:17: error: ‘QSSGRef’ was not declared in this scope
  892 |                 QSSGRef<QSSGRenderContextInterface> context;
      |                 ^~~~~~~
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:892:51: error: expected primary-expression before ‘>’ token
  892 |                 QSSGRef<QSSGRenderContextInterface> context;
      |                                                   ^
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:892:53: error: ‘context’ was not declared in this scope; did you mean ‘connect’?
  892 |                 QSSGRef<QSSGRenderContextInterface> context;
      |                                                     ^~~~~~~
      |                                                     connect
/var/tmp/portage/dev-qt/qt-creator-10.0.1/work/qt-creator-opensource-src-10.0.1/src/tools/qml2puppet/qml2puppet/editor3d/generalhelper.cpp:896:75: error: ‘class QQuick3DSceneManager’ has no member named ‘rci’
  896 |                 context = QQuick3DObjectPrivate::get(node)->sceneManager->rci;
      |        
```
which probably need to be addressed by correcting the version constraints in (R)DEPEND

Peter-Levine:
Try with 11.0.2.

Peter-Levine:
Given that the latest qt-creator ebuild was pushed without plugin-enabling USE flags, it seems that the previous resistance to such an approach (eg., https://github.com/gentoo/gentoo/pull/26675#issuecomment-1208695503) has been reversed.

Closing.

PR #255, opened 2022-06-18
dev-qt/qtdeclarative: Filter -Os

a17r:
Bug: https://bugs.gentoo.org/840861
Package-Manager: Portage-3.0.30, Repoman-3.0.3

### DIFF
--- a/dev-qt/qtdeclarative/qtdeclarative-5.15.5.9999.ebuild
+++ b/dev-qt/qtdeclarative/qtdeclarative-5.15.5.9999.ebuild
@@ -4,7 +4,7 @@
 EAPI=8
 
 PYTHON_COMPAT=( python3_{8..10} )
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
Unfortunate but it is what it is. I'd rather keep things working than be idealistic.

dev-qt/qtdeclarative/qtdeclarative-5.15.5.9999.ebuild
```diff
@@ -47,6 +47,8 @@ src_prepare() {
 }
 
 src_configure() {
+	filter-flags "-Os" # bug 840861
```
Pesa, line 50: ok, but shouldn't we replace with `-O2` instead of just filtering? I might be totally missing something, but without any optimization flags the default is `-O0` which is not a good idea.

PR #257, opened 2022-07-31
app-text/goldendict: update ebuild

mgcyung:


a17r:
It should probably be cleaned up, since dependency on qtwebkit remains unchanged, so it is impossible to build.

mgcyung:
Is there a conflict between qtwebkit and goldendict?

mgcyung:
Ok, I see.

PR #259, opened 2022-09-08
Qt 5.15.6 preparations

a17r:
Content of patchset rev 2:
https://invent.kde.org/asturmlechner/qtbase/-/commit/aa14026ac393dc9c2545bc9784104b6654255532
https://invent.kde.org/asturmlechner/qtbase/-/commit/c6f6d93ca6a1ac4e46ba7d83d13872a9b41cf95d
https://invent.kde.org/asturmlechner/qtbase/-/commit/cd0a79a5d47b7117fb30c2d04710010c2de62bc9
https://invent.kde.org/asturmlechner/qtbase/-/commit/0dcab42492c5ce61e6bfb8d30594eebb80ef6895

Added:
https://invent.kde.org/asturmlechner/qtbase/-/commit/a36d0d2565f68696ae8c81550e58d08846e12466 (fix musl w/ libexecinfo)
https://invent.kde.org/asturmlechner/qtbase/-/commit/0b746bd35143dc7f463c7995212e4dd76e5a7c08 (Remove quoting around $MAKE call)

Dropped:
Qt5BasicConfig.cmake.in: Fix cmake macro compat. for upgrade from <Qt-5.14
https://invent.kde.org/asturmlechner/qtbase/-/commit/b7b3725318f8798aa96228fc5b36b9cac06b6669
Qt 5.13 was dropped from Gentoo more than 2 years ago thus upgrades are unsupported.


thesamesam (approved):

dev-qt/qt3d/qt3d-5.15.6.9999.ebuild
```diff
@@ -3,14 +3,14 @@
 
 EAPI=8
 
+if [[ ${PV} != *9999* ]]; then
+	QT5_KDEPATCHSET_REV=1
+	KEYWORDS="~amd64 ~arm ~arm64 ~riscv ~x86"
+fi
```
Pesa, line 9: Not against this change, but I'm not sure I understand the reason (commenting on a random place, the comment refers to the whole commit)

Pesa (approved):

dev-qt/qt3d/qt3d-5.15.6.9999.ebuild
a17r: Reason is less churn when spinning new releases from live ebuilds. Packages currently using a QT5_KDEPATCHSET_REV are highly unlikely to not do so on subsequent release(s). The only module with a single patch right now is Qt5Tools, and it will only be part of a qt-5.15.x-lts tag after 2023-06-16 at least.
This is similar to KEYWORDS that are kept in sync as well (except that we won't keep actual value in sync but let it remain set to 1).

a17r:
eclass change and re-manifesting will be part of Qt 5.15.6 bump merge.

PR #265, opened 2023-02-12
x11-misc/sddm: Change RUNTIME_DIR path to "/run/sddm"

RobertBuhren:
The systemd tmpfiles config provided with sddm sets the permission for RUNTIME_DIR to 0711. Setting RUNTIME_DIR to "/run" results in the wrong permission set for this directory leading to bugs similar to:

https://github.com/systemd/systemd/issues/21009


a17r:
Thanks for your fix, unnecessary systemd issue link scrubbed.

PR #267, opened 2023-03-22
x11-misc/sddm: Outsource Gentoo defaults and resulting optional RDEPENDs to gui-apps/sddm-gentoo-config

a17r:
methods, category and package name up for discussion.

### DIFF
--- a/gui-apps/sddm-gentoo-config/sddm-gentoo-config-0.ebuild
+++ b/gui-apps/sddm-gentoo-config/sddm-gentoo-config-0.ebuild
@@ -1,0 +1,35 @@
+# Copyright 2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=8
+
+DESCRIPTION="Gentoo defaults for SDDM (Simple Desktop Display Manager)"
+HOMEPAGE="https://github.com/sddm/sddm"
+SRC_URI=""
+S=${WORKDIR}
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~loong ~ppc64 ~riscv ~x86"
+IUSE="X"
+
+RDEPEND="
+	X? ( x11-base/xorg-server )
+	!X? ( dev-libs/weston )
+"
+
+src_install() {
+	touch "${T}"/01gentoo.conf || die
+
+cat <<-EOF >> "${T}"/01gentoo.conf
+[General]
+# Which display server should be used
+DisplayServer=$(usex X "x11" "wayland")
+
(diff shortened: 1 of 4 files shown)


ArsenArsen (approved):
LGTM - can we move sddm into this category too?  :)

thesamesam (approved):
I like it and it's a UX improvement for when 0.20 lands. Thanks.

Chiitoo (approved):
Thanks!

a17r:
@all: We could also add support for 'x11-user' setting, but I haven't looked at what it would entail. Suggestions for integrating IUSE-wise welcome.

a17r:
> LGTM - can we move sddm into this category too? :)

It might be moved to `kde-plasma/` instead. https://invent.kde.org/plasma/plasma-desktop/-/issues/91

gui-apps/sddm-gentoo-config/sddm-gentoo-config-0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+S=${WORKDIR}
+
+LICENSE="public-domain"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~arm64 ~loong ~ppc64 ~riscv ~x86"
+IUSE="X"
+
+RDEPEND="
+	X? ( x11-base/xorg-server )
+	!X? ( dev-libs/weston )
```
WhatAmISupposedToPutHere, line 18: can this instead be turned into || (weston kwin) ? that way there is no unnecessary dependency on weston if kwin is installed

Chiitoo:
Is this still relevant/up to date?

a17r:
Eventually, it should become relevant, yes.

gui-apps/sddm-gentoo-config/sddm-gentoo-config-0.ebuild
```diff
@@ -0,0 +1,35 @@
...
+"
+
+src_install() {
+	touch "${T}"/01gentoo.conf || die
+
+cat <<-EOF >> "${T}"/01gentoo.conf
+[General]
+# Which display server should be used
+DisplayServer=$(usex X "x11" "wayland")
```
jospezial, line 27: I added ```DisplayServer="wayland"``` to my config manually myself.
```(WW) DAEMON: ""wayland"" is an invalid value for General.DisplayServer: fall back to "x11"```
So the value may not have ```" "``` in config.

ArsenArsen: it won't - those are shell quotes that will be discarded while lexing the arguments to `usex`

PR #269, opened 2023-06-19
dev-qt/qtnetwork: Drop bearer plugins (IUSE=connman,networkmanager)

a17r:
Upstream in #kde-devel:
- the bearer plugins were created in another era: for Symbian, when we weren't always connected to the Internet and the user maybe didn't want to turn on their data
- now, we're always connected, so the bearer plugins are unnecessary
- they've been removed from Qt 6 due to their design problems
- so just do the same in Qt 5: don't use them
- especially when upstream has already declared "the use of this API is a design flaw"

Can not drop src/plugins/bearer/generic (builds libqgenericbearer.so) though, as without it things like akonadi (KMail remains offline) and Plasma POTD stop working.

QTBUG: https://bugreports.qt.io/browse/QTBUG-114666

### DIFF
--- a/dev-qt/qtnetwork/metadata.xml
+++ b/dev-qt/qtnetwork/metadata.xml
@@ -6,12 +6,9 @@
 		<name>Gentoo Qt Project</name>
 	</maintainer>
 	<use>
-		<flag name="connman">Enable <pkg>net-misc/connman</pkg>-based bearer plugin</flag>
 		<flag name="gssapi">Enable support for GSSAPI (<pkg>virtual/krb5</pkg>)</flag>
 		<flag name="libproxy">Use <pkg>net-libs/libproxy</pkg> for automatic
 			HTTP/SOCKS proxy configuration</flag>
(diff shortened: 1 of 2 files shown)


Pesa:
Hmm the only real reason to drop them is that they're causing the issue described in the linked QTBUG, right? None of the other bullet points is particularly convincing IMHO if it wasn't for the linked bug. The bug elevates the issue from simply "the bearer plugins are useless" to "the bearer plugins are actually harmful".

Pesa:
Anyway, no objection from me other than highlighting the actual bug a bit more in the commit message.

Maybe start by masking the flags and see if anyone complains?

a17r:
Thanks, it started out with completely dropping bearer/generic which would have made ebuild changes necessary, but now it would work via masking the flags as well.

I would still do it like this, introduce it to ::gentoo via revbump so every ~arch user gets it regardless of using -N to update world, for maximum exposure, and we can keep -r0 for a while (but no longer than 5.15.10 stabilisation).

Chiitoo (approved):

Pesa (approved):

PR #270, opened 2023-08-20
dev-qt/qtcore: Drop IUSE=old-kernel

a17r:


heroxbd:
Is it because the configure options no longer work, or we had decided to drop support of the old kernels?

a17r:
The oldest kernel version you can get in Portage is `4.14.315`

Chiitoo (approved):
Thank you!

Pesa (approved):

heroxbd:
> The oldest kernel version you can get in Portage is `4.14.315`

If the kernel is managed by Gentoo, yes, we don't need `USE=old-kernel`

That said, there are cases where portage does not control the kernel, e.g. WSL1, OpenVZ, Gentoo Prefix. Users could get stuck with an old kernel.

The USE flag was introduced in https://bugs.gentoo.org/669994 and in https://github.com/gentoo/qt/pull/214 

It is still needed and being used in various Gentoo scenarios.

a17r:
Right, but that was more than 3 years ago. Is there an overview which distributions are still shipping with older kernels than 4.14?

heroxbd:
Andreas Sturmlechner ***@***.***> writes:

> Right, but that was more than 3 years ago. Is there an overview which
> distributions are still shipping with older kernels than 4.14?

Yes, for example, CentOS 7 with linux-3.10 (EOL mid 2026) is still in
use on many HPC sites.

a17r:
Hm. I'm not sure we should wait *that* long but there's no hurry right now either.

heroxbd:
Andreas Sturmlechner ***@***.***> writes:

> Hm. I'm not sure we should wait that long but there's no hurry right now either.

Got you, Andreas.  Thanks for your consideration.

PR #274, opened 2024-01-01
qt5-build.eclass: filter-lto earlier for qtbase module packages

a17r:
Bug: https://bugs.gentoo.org/919789

### DIFF
--- a/eclass/qt5-build.eclass
+++ b/eclass/qt5-build.eclass
@@ -179,6 +179,15 @@ fi
 qt5-build_src_prepare() {
 	qt5_prepare_env
 
+	# Workaround for bug #915203
+	# Upstream: https://bugreports.qt.io/browse/QTBUG-111514
+	if [[ ${PN} != qtcore ]]; then
+		append-ldflags $(test-flags-CCLD -Wl,--undefined-version)
+	fi
+
+	# many bugs, no one to fix (already done for qtbase)
+	filter-lto
+
 	if [[ ${QT5_BUILD_TYPE} == live ]] || [[ -n ${KDE_ORG_COMMIT} ]]; then
 		if [[ -n ${KDE_ORG_COMMIT} ]]; then
 			einfo "Preparing KDE Qt5PatchCollection snapshot at ${KDE_ORG_COMMIT}"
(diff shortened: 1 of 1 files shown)


thesamesam (approved):

eclass/qt5-build.eclass
```diff
@@ -226,6 +229,9 @@ qt5-build_src_prepare() {
 qt5-build_src_configure() {
 	if [[ ${QT5_MODULE} == qtbase ]]; then
 		qt5_base_configure
+	else
+		# many bugs, no one to fix (already done for qtbase)
+		filter-lto
```
Pesa, line 234: is there a reason we can't do this for every module at the beginning of `qt5-build_src_prepare`?

ionenwks (approved):

eclass/qt5-build.eclass
a17r: we could, these things are only usually done in src_configure and the things we do to qmake buildsys is the outlier here...

ionenwks: If I had to pick I'd rather no duplication over something minor like this, not that I mind much either way.

But just remembered that I actually used src_prepare in qt6-build.eclass for another reason e.g. for ebuilds doing:
```
src_configure() {
    *does something with flags*
    qt6-build_src_configure
}
```
And I wanted src_configure to have the flags pre-modified from the eclass at that stage -- notably important for the qtbase hacks with -m* flags the eclass changes. So qt6-build.eclass' filter-lto is in src_prepare right now.

ionenwks: Not that qt6-build's reasons matter for qt5-build, but for what it's worth qt5-build using prepare would consistent with what qt6-build does :shrug:

ionenwks (approved):

Chiitoo (approved):
Thank you!

eclass/qt5-build.eclass
Pesa: Yeah I'm not a fan of the duplication either (although I admit it's pretty minor). Also, doing it earlier may be more robust against future changes.

Pesa (approved):
Thanks!

PR #275, opened 2024-01-02
dev-qt/qtcore - add wall of blockers against breakage

a17r:
Revdeps can drop dev-qt/* dependencies on version bumps at any time, which then
may lead to Portage disregarding these dev-qt/* packages for upgrade if nothing
is left depending on them. When these packages contain Qt5 plugins it breaks
runtime hard - broken screenlocker, blank login manager, ... -, with:

`Cannot mix incompatible Qt library (5.15.10) with this library (5.15.11)`

We've had this happen with dev-qt/qtvirtualkeyboard and dev-qt/qtcharts already
and gradual porting to Qt6 ensures it will happen more often in the future.

For lack of a better mechanism from the package manager, this blocker package
makes sure dep calculation will take care of the necessary cleanup of dangling
Qt packages, even if it will not improve our backtracking situation.

### DIFF
--- a/dev-qt/qtcore/qtcore-5.15.12.9999.ebuild
+++ b/dev-qt/qtcore/qtcore-5.15.12.9999.ebuild
@@ -24,7 +24,50 @@ DEPEND="
 	icu? ( dev-libs/icu:= )
 	!icu? ( virtual/libiconv )
 "
-RDEPEND="${DEPEND}"
+RDEPEND="${DEPEND}
+	!<dev-qt/designer-${QT5_PV}:5
+	!<dev-qt/qt3d-${QT5_PV}:5
+	!<dev-qt/qtbluetooth-${QT5_PV}:5
+	!<dev-qt/qtcharts-${QT5_PV}:5
+	!<dev-qt/qtconcurrent-${QT5_PV}:5
+	!<dev-qt/qtdatavis3d-${QT5_PV}:5
+	!<dev-qt/qtdbus-${QT5_PV}:5
+	!<dev-qt/qtdeclarative-${QT5_PV}:5
+	!<dev-qt/qtgamepad-${QT5_PV}:5
+	!<dev-qt/qtgraphicaleffects-${QT5_PV}:5
+	!<dev-qt/qtgui-${QT5_PV}:5
+	!<dev-qt/qthelp-${QT5_PV}:5
+	!<dev-qt/qtimageformats-${QT5_PV}:5
+	!<dev-qt/qtlocation-${QT5_PV}:5
+	!<dev-qt/qtmultimedia-${QT5_PV}:5
+	!<dev-qt/qtnetwork-${QT5_PV}:5
+	!<dev-qt/qtnetworkauth-${QT5_PV}:5
+	!<dev-qt/qtopengl-${QT5_PV}:5
+	!<dev-qt/qtpositioning-${QT5_PV}:5
+	!<dev-qt/qtprintsupport-${QT5_PV}:5
+	!<dev-qt/qtquickcontrols-${QT5_PV}:5
+	!<dev-qt/qtquickcontrols2-${QT5_PV}:5
+	!<dev-qt/qtquicktimeline-${QT5_PV}:5
+	!<dev-qt/qtscript-${QT5_PV}:5
+	!<dev-qt/qtscxml-${QT5_PV}:5
+	!<dev-qt/qtsensors-${QT5_PV}:5
+	!<dev-qt/qtserialbus-${QT5_PV}:5
+	!<dev-qt/qtserialport-${QT5_PV}:5
+	!<dev-qt/qtspeech-${QT5_PV}:5
+	!<dev-qt/qtsql-${QT5_PV}:5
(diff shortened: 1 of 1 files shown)


thesamesam (approved):
I'm worried about this not working so well in the wild once we actually try it out, but I'm okay with trying it - just reserve the right to revisit ofc.

Please also perhaps explicitly note in the commit message why https://bugs.gentoo.org/918873 wouldn't help here.

My preference would be that we don't do this for the upcoming stablereq for 5.15.12 and instead do it on a version we will let soak in ~arch.

ionenwks:
Maybe virtual/ would be better? I don't want this to sound like it provides anything (like users wondering if there's a new qtenforcement Qt module or something). Then again, dev-qt may work out better for people doing `dev-qt/*` accept keywords and not having to worry about a separate package, so I dunno.

Guess could consider to just put it in the RDEPEND of the eclass directly (not sure if this makes anything worse except maybe portage having to work harder), maybe be harder to control given changes require bumping the entire stack.

Do wonder how this will really behave anyway (haven't looked too closely at how portage handles softblocks). Kind of worry of what happens if e.g. new version qtenforcement is emerged early, then users get a build failure (perhaps another package in-between enforcement and qtcore) and emerge aborts, are all old-version dev-qt/* packages gone from their system at this point? Are users stuck without a working DE until they fix things?

ionenwks:
> Do wonder how this will really behave anyway (haven't looked too closely at how portage handles softblocks). Kind of worry of what happens if e.g. new version qtenforcement is emerged early, then users get a build failure (perhaps another package in-between enforcement and qtcore) and emerge aborts, are all old-version dev-qt/* packages gone from their system at this point? Are users stuck without a working DE until they fix things?

Being in the eclass directly may at least ensure dev-qt/qtcore was successfully merged before it removes everything else.

ionenwks:
Guess the blocker list could be reduced too to only those that can actually cause problems too. Not that I kept up with this issue too closely.

ionenwks:
> > Do wonder how this will really behave anyway (haven't looked too closely at how portage handles softblocks). Kind of worry of what happens if e.g. new version qtenforcement is emerged early, then users get a build failure (perhaps another package in-between enforcement and qtcore) and emerge aborts, are all old-version dev-qt/* packages gone from their system at this point? Are users stuck without a working DE until they fix things?
> 
> Being in the eclass directly may at least ensure dev-qt/qtcore was successfully merged before it removes everything else.

Actually, is there a reason to having these blockers everywhere except just qtcore?

a17r:
> Actually, is there a reason to having these blockers everywhere rather than just in qtcore?
Back when this was first proposed we were mainly dealing with portage allowing half-Qt5-upgrades thus breaking users' systems, so I went for the ultimate option. I tested it upgrading sucessfully back then, dunno if it changes Portage's depgraph for better or worse if just in dev-qt/qtcore - but yes, much easier how you suggest thus better.

ionenwks (approved):
That feels better, as horrible as it still is but I don't have a better solution to offer.

Guess Qt6 should do this too? (I can add it later myself, assuming nothing blew up)

Pesa (approved):
I don't like it but this is what we had to do many years ago with Qt4 as well. If there is no better solution nowadays, I guess we'll have to accept the wall of blockers...

a17r:
> Guess the blocker list could be reduced to only those that can actually cause problems too. Not that I kept up with this issue too closely.

That list would be difficult to assemble.

```
# equery b /usr/lib64/qt5/plugins | grep dev-qt
dev-qt/designer-5.15.12
dev-qt/qt3d-5.15.12
dev-qt/qtdeclarative-5.15.12
dev-qt/qtgui-5.15.12
dev-qt/qtimageformats-5.15.12
dev-qt/qtlocation-5.15.12
dev-qt/qtmultimedia-5.15.12
dev-qt/qtnetwork-5.15.12
dev-qt/qtpositioning-5.15.12
dev-qt/qtprintsupport-5.15.12
dev-qt/qtsensors-5.15.12
dev-qt/qtserialbus-5.15.12
dev-qt/qtsql-5.15.12
dev-qt/qtsvg-5.15.12
dev-qt/qtvirtualkeyboard-5.15.12
dev-qt/qtwayland-5.15.12
dev-qt/qtwebview-5.15.12
dev-qt/qtwidgets-5.15.12
```

That does not show qtcharts even though I know it broke things on one of the systems I manage...

Chiitoo (approved):

ionenwks:
> That does not show qtcharts even though I know it broke things on one of the systems I manage...

Ah well, and guess the list is rather long either way... may as well stick with the whole deal then.

a17r:
Packages we could drop for sure from there:

```
dev-qt/assistant:0
dev-qt/linguist:0
dev-qt/linguist-tools:0
dev-qt/pixeltool:0
dev-qt/qdbus:0
dev-qt/qdbusviewer:0
dev-qt/qdoc:5
dev-qt/qtdiag:5
dev-qt/qtpaths:5
dev-qt/qtplugininfo:5
dev-qt/qttranslations:5
dev-qt/qtwaylandscanner:5
```

PR #278, opened 2024-03-06
x11-misc/sddm-9999: PAM has been removed upstream include files manually

FireBurn:
https://github.com/sddm/sddm/commit/ae072f901671b68861da9577e3e12e350a9053d5

PAM modules are distribution specific and are ultimately better fit in the distribution packaging.

Having them in SDDM brings extra burden to the already stretched developers and maintainers. Plus a handful of distributions currently patch the (Arch) PAM modules, where having the full set in their own tree would make for cleaner solution.

I've added the pam files that uses to be installed into the files directory and manually installed them, this keeps things working. Not sure if they should be versioned or not


x11-misc/sddm/sddm-9999.ebuild
```diff
@@ -68,9 +68,6 @@ PATCHES=(
 	# Downstream patches
 	"${FILESDIR}/${PN}-0.20.0-respect-user-flags.patch"
 	"${FILESDIR}/${PN}-0.21.0-Xsession.patch" # bug 611210
-	"${FILESDIR}/${PN}-0.20.0-sddm.pam-use-substack.patch" # bug 728550
-	"${FILESDIR}/${PN}-0.21.0-disable-etc-debian-check.patch"
-	"${FILESDIR}/${PN}-0.21.0-no-default-pam_systemd-module.patch" # bug 669980
```
a17r, line 73: simply dropping this patch is re-introducing bug 669980.

a17r (changes requested):

